Issue #1, opened 2024-12-27
[beman-tidy] Propose/create bemanification script/framework


neatudarius:
Propose/create bemanification script

- [x] Proposed name: `beman-tidy` - as discussed with @dietmarkuehl in November 2024 in ISO C++ Meeting - Poland
- [x] Build PoC
- [x] Demo and get feedback
- [x] beman-tidy MVP

bretbrownjr:
For `beman-tidy` I would start with *really* basic features. Such as:

* Does the project have a README, CONTRIBUTING, and LICENSE?

* Does the project have a CMakeLists.txt next to the README?

* Can we validate other project structure basics that are already in The Beman Standard?

Just packaging and delivering the tool to the existing Beman libraries is interesting enough work, I think.

Examples of more complicated features that need some dependencies and/or consensus established:

* Building the code

* Meaningful static analysis 

* Simple static analysis such as simple text search through files

* Complex static analysis such as clang-based parsing (this at least requires a successful CMake invitation to get a compile commands JSON file)

phelter:
Not sure if this helps but I've open sourced my own repo : https://github.com/Retlek-Systems-Inc/rs_cmake that does most of the heavy lifting of the above.

I'm in the process of making the automation easier for various different CI/CD chains using CMakePresets.

* Supports CMake building the code.
* Supports easy way for creating and identifying Mocks and Test frameworks by replacing the implementation with fake/mock models instead.
* Pedantic error - opt-out instead of opt-in model for all builds and checks.
* Common Open-source static analysis (clang-tidy, cppcheck, cpplint) - CMake functions for per-target disablement of types.
* Common Open-source dynamic sanitizers - (asan,tsan,msan, etc)
* Creation of Code coverage.
* Stored docker containers for running all CI/CD on various platforms - Github actions for CI path (missing CD portion).
   * Missing windows environments.

neatudarius:
beman-tidy MVP is ready for review - https://github.com/bemanproject/infra/pull/2 (first PR).
Also created this discourse thread: [The Codebase Bemanification Tool - beman-tidy is ready for MVP!](https://discourse.bemanproject.org/t/the-codebase-bemanification-tool-beman-tidy-is-ready-for-mvp/435)

CC: @bemanproject/core-reviewers @bemanproject/leads

neatudarius:
Implemented in https://github.com/bemanproject/infra/pull/2

PR #2, opened 2025-01-22
[beman-tidy] Add beman-tidy: The Codebase Bemanification Tool

neatudarius:
#1  - Add beman-tidy: The Codebase Bemanification Tool

`PLEASE START REVIEW WITH A FULL READ OF README.md.`
* 200 lines of docs in root README.md
* Python code: only 750 production code to read in `beman-tidy` file + `lib/` directory.

Context: This idea is co-authored with @dietmarkuehl . Skelethon started in Poland, multiple checks implemented. This PR proposes a minimal example of how the tool would be like - check `tools/beman-tidy/README.md`.

Updates proposed in this PR:
- [x] `beman-tidy` script / tool and it's usage. Check docs - please start reviewing this PR with the docs..
- [x] Docs `tools/beman-tidy/README.md`: 
    - [x] Purpose
    - [x] Iterative implementation approach and limitations
    - [x] Install instructions
    - [x] Usage instructions
    - [x] Development flow
- [x] `.beman-standard.yml`: offline STABLE snapshot of The Beman Standard (high level set of checks).
- [x] Beman Standard checks implementation examples - ~~3~~ 2 examples see `tools/beman-tidy/lib/checks/beman_standard/readme.py`.
   - [x] `README.TITLE`
   - [x] `README.BADGES`
   ~~- [x] `README.LIBRARY_STATUS`~~ Created a follow-up PR examples to add README.LIBRARY_STATUS, supposing the framework is already merged - https://github.com/neatudarius/infra/pull/7
- [x] Testing
    - [x] Testing framework
    - [x] Docs for writing new tests.
    - [x] Tests for implemented checks
    - [x] Validate portability (of the tools and the build scripts) - deferred to #25 

- [x] pre-commit integration 
  - [x] Example of using inside examplar - https://github.com/bemanproject/exemplar/pull/181/files 
 

Follow-up work (new PRs):
-  Implement 39/41 checks. Note: Some of them will be skipped in first iteration (e.g, rules related to repo settings, releases on github etc). - #27
- After `beman-tidy` has enough coverage over the Beman Standard / `beman.exemplar`, we'll promote it on all repos to be run in dry-run mode by default. - #28


tools/beman-tidy/README.md
```diff
@@ -0,0 +1,75 @@
...
+# no errors found example
+$ ./tools/beman-tidy/beman-tidy --repo_path ../exemplar
+Checks pipeline started ...
+
+Running check [REQUIREMENT][FIX_INPLACE_INCOMPATIBLE_WITH_UNSTAGED_CHANGES] ... 
+ check [REQUIREMENT][FIX_INPLACE_INCOMPATIBLE_WITH_UNSTAGED_CHANGES] ... PASSED
+
+...
+
+Checks pipeline completed.
```
RaduNichita: Maybe we should put an explicit message that no errors were found?

tools/beman-tidy/README.md
```diff
@@ -0,0 +1,75 @@
...
+## Usage
+
+TODO: Add a usage section.
+
+```shell
+$ ./tools/beman-tidy/beman-tidy --help
+  -h, --help            show this help message and exit
+  --repo_path REPO_PATH
+                        path to the repository to check
+  --fix_inplace, --no-fix_inplace
```
RaduNichita: I think it's better to have `--dry-run` that does not modify anything.

tools/beman-tidy/README.md
```diff
@@ -0,0 +1,75 @@
...
+TODO: Add an installation section.
+
+## Usage
+
+TODO: Add a usage section.
+
+```shell
+$ ./tools/beman-tidy/beman-tidy --help
+  -h, --help            show this help message and exit
+  --repo_path REPO_PATH
```
RaduNichita:
```suggestion
  --repo_path <REPO_PATH>
```

neatudarius: That help menu it's autogenerated by argparse. Cannot be changed.

tools/beman-tidy/README.md
neatudarius: done

tools/beman-tidy/README.md
neatudarius: done.

wusatosi (commented):
This should be trivial to integrate with pre-commit right?
I don't see any `.pre-commit-hooks.yaml` files.
See an example here: https://github.com/pre-commit/pre-commit-hooks/blob/main/.pre-commit-hooks.yaml

tools/beman-tidy/images/usage-example.png
RaduNichita, line 1: Why do we need a photo and not simply put thes coomand and their output?

tools/beman-tidy/lib/checks/base/base_check.py
```diff
@@ -0,0 +1,94 @@
...
+        self.repo_info = repo_info
+        assert "name" in repo_info
+        self.repo_name = repo_info["name"]
+        assert "top_level" in repo_info
+        self.repo_path = repo_info["top_level"]
+        assert self.repo_path is not None
+        self.library_name = f"beman.{self.repo_name}"
+        assert self.library_name is not None
+
+    def default_check(self):
```
RaduNichita: The name of this function is misleading imo

Can you add something to show that this class is only internally used?

tools/beman-tidy/lib/checks/base/base_check.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def fix(self):
+        """
+        Fixes the issue if the Beman Standard is not applied.
+        - If check already applied, this method is a no-op and should return True.
+        - Otherwise, it will try to apply the check inplace. Returns the status of the fix attempt.
+        """
+        return False
+
+    def log(self, message, enabled=True):
```
RaduNichita, line 101: Why not using this library for logging instead?

https://docs.python.org/3/library/logging.html

tools/beman-tidy/lib/checks/base/file_base_check.py
```diff
@@ -0,0 +1,105 @@
...
+        """
+        Read the file content as lines.
+        """
+        try:
+            with open(self.path, 'r') as file:
+                return file.readlines()
+        except Exception as e:
+            return []
+
+    def read_lines_strip(self):
```
RaduNichita, line 78: I think we can have only one method of reading from file, I find it as unnecessary logic to have duplicate methods with more or less the same logic

tools/beman-tidy/lib/checks/base/file_base_check.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def replace_line(self, line_number, new_line):
+        """
+        Replace the line at the given line number with the new line.
+        """
+        lines = self.read_lines()
+        lines[line_number] = new_line
+        self.write_lines(lines)
+
+    def is_empty(self):
```
RaduNichita, line 108: I think we can do this from file metadata without reading the file.

tools/beman-tidy/lib/checks/beman_standard/readme.py
```diff
@@ -0,0 +1,93 @@
...
+class ReadmeBadgesCheck(ReadmeBaseCheck):
+    def __init__(self, repo_info, beman_standard_check_config):
+        super().__init__(repo_info, beman_standard_check_config)
+
+    def check(self):
+        """
+        self.config["values"] contains a fixed set of Beman badges.
+        """
+        badges = self.config["values"]
+        assert len(badges) == 4  # The number of library maturity model states
```
RaduNichita, line 57: Can we put this constant somewhere in a config or as a constant in class?

tools/beman-tidy/lib/checks/beman_standard/readme.py
```diff
@@ -0,0 +1,93 @@
...
+
+        # Check if at exactly one of the required badges is present.
```
RaduNichita:
```suggestion
        # Check if exactly one of the required badges is present.
```

tools/beman-tidy/lib/pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+from .checks.beman_standard.cpp import *
+from .checks.beman_standard.directory import *
+from .checks.beman_standard.file import *
+from .checks.beman_standard.general import *
+from .checks.beman_standard.license import *
+from .checks.beman_standard.readme import *
+from .checks.beman_standard.release import *
+from .checks.beman_standard.toplevel import *
+
+red_color = "\033[91m"
```
RaduNichita, line 20: I would suggest using the default python class suggested - https://docs.python.org/3/library/logging.html

tools/beman-tidy/lib/pipeline.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/python3
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from .checks.system.registry import *
```
RaduNichita:
```suggestion
from .checks.system.registry import *
```

Can we import only what we only need from that file? Usually to import * is a bad practice.

tools/beman-tidy/lib/pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+from .checks.beman_standard.release import *
+from .checks.beman_standard.toplevel import *
+
+red_color = "\033[91m"
+green_color = "\033[92m"
+yellow_color = "\033[93m"
+gray_color = "\033[90m"
+no_color = "\033[0m"
+
+
```
RaduNichita, line 26:
```suggestion
```

tools/beman-tidy/lib/utils/git.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/python3
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import os
+import sys
+import yaml
+
+from git import Repo, InvalidGitRepositoryError
+
+
```
RaduNichita, line 9:
```suggestion
```

tools/beman-tidy/lib/utils/git.py
```diff
@@ -0,0 +1,115 @@
...
+        # Get the top-level directory of the repository
+        top_level_dir = repo.git.rev_parse("--show-toplevel")
+
+        # Get the repository name (directory name of the top level)
+        repo_name = os.path.basename(top_level_dir)
+
+        # Get the remote URL (assuming 'origin' is the remote name)
+        remote_url = None
+        if "origin" in repo.remotes:
```
RaduNichita, line 28:
```suggestion
        if "origin" in repo.remotes:
```

Can we make it customizable?

tools/beman-tidy/tests/__init__.py
RaduNichita, line 1: Do we need this empty file? https://docs.python.org/3/reference/import.html#regular-packages

RaduNichita (commented):
It looks good, I did a first pass and added some comments.

However, I suggest splitting it into smaller PRs - I would recommend first merging this PR with 1 check, then do other follow-up PRs - one PR per check.

tools/beman-tidy/tests/__init__.py
neatudarius: Yes. It's Python standard practice to generate a tree o deps which allows relative import from parents / parallel subtrees.

Note: This is needed from various reasons, one is that I want to have all production in `lib/` and all tests in `tests/.`.

tools/beman-tidy/lib/utils/git.py
neatudarius: I don't think we care, at least it's not part from first version in my opinion.

tools/beman-tidy/lib/utils/git.py
neatudarius: This is intended. Please check CI file. The code is automatically intended with autopep8, which by default put 2 blanks  lines after imports or between 2 functions. So I will keep current status.

PS The PR passes the linter check on CI.

tools/beman-tidy/lib/checks/base/file_base_check.py
neatudarius: For sure we need the content in multiple shapes:
- a full string with a file where for instance you match if the file has a string anywhere inside
- an array of lines when you need to correlate checks: the line after purpose line must ne empty and the next line must ne status line.

These small tweaks will have a huge impact on the simplifing the complexity in the check classes.

tools/beman-tidy/lib/checks/base/file_base_check.py
neatudarius: I don't think it's worth the effort. beman-tidy will read read multiple times a file because for a given check may also mutate the content to detect something or to fix it. I propose this simple architecture where all checks are independent  z each reads README.md, fixesit, saves it.  It will make debugging  inside this tools easier.

More reasons here: https://github.com/bemanproject/infra/pull/2#discussion_r2135187317

tools/beman-tidy/lib/pipeline.py
neatudarius: Will ignore, as explained here: https://github.com/bemanproject/infra/pull/2#discussion_r2135182463

tools/beman-tidy/images/usage-example.png
neatudarius: I have both, as the code snippets does not show the user the colors (red/green for failure and success).

tools/beman-tidy/lib/pipeline.py
neatudarius: done

tools/beman-tidy/lib/checks/base/base_check.py
neatudarius: > Can you add something to show that this class is only internally used?

I made all classes from base/ to be abstract, with abc package, and now they cannot be instantiated directly. Thus, it obvious it's for internal use, IMO.


> The name of this function is misleading imo

*Check.default_check() is a default check for a type (e..g, for a file it means the file exists etc). It's called inside the pipeline as a prestep for all checks.
If and only if default_check() passes, the specific check() logic is called. 
Note: default_check() is used only for internal purposes. When you implement a new check class, you don't care.
image

tools/beman-tidy/lib/checks/beman_standard/readme.py
neatudarius: Already put in `.beman-standard.yml`. Thanks! I applied the suggestion to use the data.

tools/beman-tidy/lib/pipeline.py
neatudarius: I will look for it - I need to loggers in this tools
- 1) print() - always log
- 2) custom log which is enabled only for verbose mode or debugging

neatudarius:
> However, I suggest splitting it into smaller PRs - I would recommend first merging this PR with 1 check, then do other follow-up PRs - one PR per check.

I will think about it. Initially I wanted the same think, but that means we'll not be able to add multiple types of tests or relevant examples in docs. The current PR is also relevant for tooling and CI flows.

More over, let's assume I remove 2/3 checks and we keep only README.TITLE. Only 2 files will change:
```
dariusn ~/dev/dn/git/Beman/infra/tools/beman-tidy [beman-tidy] $ cat lib/checks/beman_standard/readme.py| wc -l
      99
dariusn ~/dev/dn/git/Beman/infra/tools/beman-tidy [beman-tidy] $ cat tests/beman_standard/readme/*  | wc -l
cat: tests/beman_standard/readme/data: Is a directory
     153
```
And we'll remove less than 200 lines from this PR, still keeping ~2K lines in this PR. So it does not worth, IMO.

Note: Upcoming PR will be smaller, but we need this to be a solid and testable framework.

neatudarius:
> This should be trivial to integrate with pre-commit right? 

Yes, it should be, at least pre-commit can pass the path to the repo to the beman-tody tool.

>I don't see any `.pre-commit-hooks.yaml` files. See an example here: https://github.com/pre-commit/pre-commit-hooks/blob/main/.pre-commit-hooks.yaml

@wusatosi , I will try and come back. Thanks for suggestion!

tools/beman-tidy/lib/checks/base/base_check.py
RaduNichita: what about precheck or pre_check? Would that be more suitable?

tools/beman-tidy/lib/checks/base/file_base_check.py
RaduNichita: Can we have a generic function that we can use for reading? I still believe it is a bit too much to go through many similar function for such a simple operation.

tools/beman-tidy/lib/checks/base/file_base_check.py
RaduNichita: Ok, we can work on optimization later.

neatudarius: I don't think we want to do that. It would be faster, but not with isolated steps. I think I want to keep all checks running an independent flow: any README checks should read the file from disk, change it, save it, before going to the next check.

tools/beman-tidy/lib/checks/base/file_base_check.py
neatudarius: It's literally the python way, aka I just made a wrapper for each file.read* function.

tools/beman-tidy/Makefile
```diff
@@ -0,0 +1,34 @@
...
+# Install tool:
+install:
+	@echo "Installing production dependencies..."
+	pip3 install -r requirements.txt &> /dev/null
+
+# Install development dependencies:
+install-dev:
+	@echo "Installing development dependencies..."
+	pip3 install -r requirements-dev.txt &> /dev/null
+	brew install autopep8 &> /dev/null
```
rishyak, line 15: Relying on `brew` ties this to macOS and assumes users have Homebrew installed. A `pyproject.toml`-driven setup would be truly cross-platform.

tools/beman-tidy/Makefile
rishyak, line 1: The Python ecosystem now provides standardised tooling via `pyproject.toml`, so a hand-rolled Makefile isn't necessary. Let's consolidate build and dependency management there.

The same applies for the `requirements.txt` files.

rishyak (commented):
To improve cross-platform compatibility and align with modern Python packaging standards, I'd recommend moving the build and dependency definitions into a `pyproject.toml` rather than maintaining a custom Makefile and requirements.txt files.

[PEP 518](https://peps.python.org/pep-0518/) defines this declarative, tool-agnostic format, which most build backends now support out of the box. In particular, Astral's [uv](https://docs.astral.sh/uv/) provides seamless project management around `pyproject.toml`. I've been using it both professionally and in my made-for-fun CLI (see https://github.com/rishyak/tardy/blob/main/pyproject.toml).

This switch will simplify the build process, reduce maintenance overhead, and ensure everyone can build the project similarly on any platform.

Edit: See https://github.com/rishyak/infra/tree/rish.uv.bemantidy and if you like it, [here's the PR](https://github.com/neatudarius/infra/pull/4).

tools/beman-tidy/.beman-standard.yml
```diff
@@ -0,0 +1,161 @@
...
+        ]
+FILE.COPYRIGHT:
+    - type: RECOMMENDATION
+    - regex: <put a regex here if possible - exclude copyright lines>
+
+# CPP
+CPP.NAMESPACE:
+    - type: RECOMMENDATION
+CPP.NO_FLAG_FORKING:
+    - type: RECOMMENDATION
```
JeffGarland: Well @camio has created *another mess* (wink wink) here because this is now a requirement -- against my wishes lol.  So the inability to reference the actual latest beman standard means these things will constently be out of sync unless we're making sure. So I guess the comment is make it a requirement

tools/beman-tidy/README.md
```diff
@@ -0,0 +1,254 @@
...
+<!--
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
+## Description
+
+`beman-tidy` is a tool used to check and apply [The Beman Standard](https://github.com/bemanproject/beman/blob/main/docs/BEMAN_STANDARD.md).
+
+Purpose: The tool is used to `check` (`--dry-run`) and `apply` (`--fix-inplace`) the Beman Standard to a repository.
+Note: `07.06.2025`: In order to make the best and quickly use of the tool in the entire organization, most of the checks will not support the `--fix-inplace` flag in the first iteration.
```
JeffGarland:
```suggestion
Note: `2025-06-07`: In order to make the best and quickly use of the tool in the entire organization, most of the checks will not support the `--fix-inplace` flag in the first iteration.
```

ISO times please -- the only rational and universal way

tools/beman-tidy/README.md
```diff
@@ -0,0 +1,254 @@
...
+* Find a Beman Standard check that is not implemented.
+* Add a new entry to the `.beman-standard.yml` file.
+* Add a new check to the `lib/checks/beman_standard/` directory (find existing checks for inspiration).
+* Add tests for the new check.
+* Run the tests.
+* Commit the changes.
+
+Requirements:
+* `beman-tidy` must be able to run on Windows, Linux, and macOS, thus it's 100% Python.
+* `beman-tidy` must NOT used internet access.  A local snapshot of the standard is used (check `.beman-standard.yml`).
```
JeffGarland:
```suggestion
* `beman-tidy` must NOT use internet access.  A local snapshot of the standard is used (check `.beman-standard.yml`).
```

tools/beman-tidy/README.md
```diff
@@ -0,0 +1,254 @@
...
+* `beman-tidy` must have `verbose` and `non-verbose` modes. Default is `non-verbose`.
+* `beman-tidy` must have `dry-run` and `fix-inplace` modes. Default is `dry-run`.
+* `beman-tidy` must detect types of checks: failed, passed, skipped (not implemented) and print the summary/coverage.
+
+Limitations:
+* `07.06.2025`: `beman-tidy` will not support the `--fix-inplace` flag in the first iteration for most of the checks.
```
JeffGarland:
```suggestion
* `2025-06-07`: `beman-tidy` will not support the `--fix-inplace` flag in the first iteration for most of the checks.
```
iso times please

tools/beman-tidy/README.md
```diff
@@ -0,0 +1,254 @@
...
+* `07.06.2025`: `beman-tidy` may generate small changes to the standard (e.g., for automated fixes), while the standard is not stable. Thus, the tool itself may be unstable.
```
JeffGarland:
```suggestion
* `2025-06-07`: `beman-tidy` may generate small changes to the standard (e.g., for automated fixes), while the standard is not stable. Thus, the tool itself may be unstable.
```

tools/beman-tidy/README.md
```diff
@@ -0,0 +1,254 @@
...
+        └── conftest.py
+
+24 directories, 62 files
+```
+
+Notes:
+
+* `beman-tidy`: A Python script that is used to check and apply the Beman Standard to a repository.
+* `.beman-standard.yml`: Stable version of the standard; the tool does not fetch the latest unstable version of the standard.
+* `lib/`: The library for the beman-tidy tool (e.g, checks, utils, etc.).
```
JeffGarland: so we now have the beman-module concept in infra.  Should we consider using that to help allowing internet fetch of the latest standard?

tools/beman-tidy/.beman-standard.yml
neatudarius: Hello @JeffGarland , I want to better understand the problems:

P1: Some of the rule may be changed in latest standard https://github.com/bemanproject/beman/blob/main/docs/BEMAN_STANDARD.md#cppno_flag_forking, and I need to update the `.beman-standard.yml` config. No comment, I will do that!


P2: The tool does not DOWNLOAD BEMAN_STANDARD.md from beman/ repo. Initially I did that, but Beman folks gave feedback in Austria to be completely and offline tool.
Is P2 a problem for you?

P3. We can still fetch a copy of BEMAN_STANDARD.md inside this repo, with beman-module, as suggested by you here https://github.com/bemanproject/infra/pull/2#discussion_r2138996622. 
P3.1. We can reference to the local copy (e.g. console logs which states that the implemented rule is REQUIREMENT or RECOMMANDATIN, for this SHA).
P3.2. I don't think implementing P3.1. removes the need for `.beman-tidy.yml` config. Ofc, we can remove the config, but we'll move hardcoded data and regexes into the Python code AND probably the complexity of the Python code will increase.

@JeffGarland , please check.

tools/beman-tidy/README.md
neatudarius: Moved this thread in https://github.com/bemanproject/infra/pull/2#discussion_r2140997573.

tools/beman-tidy/lib/checks/base/base_check.py
neatudarius: I will integrate this feedback.

.github/workflows/beman-tidy.yml
```diff
@@ -0,0 +1,31 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+name: Run Beman Tidy Tests
```
RaduNichita:
```suggestion
name: Run Beman Tidy Checks
```

tools/beman-tidy/README.md
```diff
@@ -0,0 +1,254 @@
...
+
+<!--
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
+## Description
+
+`beman-tidy` is a tool used to check and apply [The Beman Standard](https://github.com/bemanproject/beman/blob/main/docs/BEMAN_STANDARD.md).
+
+Purpose: The tool is used to `check` (`--dry-run`) and `apply` (`--fix-inplace`) the Beman Standard to a repository.
```
RaduNichita, line 11: What do you think about not having this opting of fix-inplace in the first version.

I think it's of no use to have the parameter if we don't support at least 10-20 checks

tools/beman-tidy/README.md
```diff
@@ -0,0 +1,254 @@
...
+Expected Development Flow:
+
+* Find a Beman Standard check that is not implemented.
+* Add a new entry to the `.beman-standard.yml` file.
+* Add a new check to the `lib/checks/beman_standard/` directory (find existing checks for inspiration).
+* Add tests for the new check.
+* Run the tests.
+* Commit the changes.
+
+Requirements:
```
RaduNichita: Why do we need this section, it is something that I would expect to be in a design doc, not in a README

tools/beman-tidy/README.md
```diff
@@ -0,0 +1,254 @@
...
+* `beman-tidy` must detect types of checks: failed, passed, skipped (not implemented) and print the summary/coverage.
+
+Limitations:
+* `2025-06-07`: `beman-tidy` will not support the `--fix-inplace` flag in the first iteration for most of the checks.
+* `2025-06-07`: `beman-tidy` may generate small changes to the standard (e.g., for automated fixes), while the standard is not stable. Thus, the tool itself may be unstable.
+
+### Tree structure
+
+```shell
+ $ tree .
```
RaduNichita: I would only put the relevant things, not all files.

tools/beman-tidy/README.md
```diff
@@ -0,0 +1,254 @@
...
+* `beman-tidy` must NOT use internet access.  A local snapshot of the standard is used (check `.beman-standard.yml`).
+* `beman-tidy` must have `verbose` and `non-verbose` modes. Default is `non-verbose`.
+* `beman-tidy` must have `dry-run` and `fix-inplace` modes. Default is `dry-run`.
+* `beman-tidy` must detect types of checks: failed, passed, skipped (not implemented) and print the summary/coverage.
+
+Limitations:
+* `2025-06-07`: `beman-tidy` will not support the `--fix-inplace` flag in the first iteration for most of the checks.
+* `2025-06-07`: `beman-tidy` may generate small changes to the standard (e.g., for automated fixes), while the standard is not stable. Thus, the tool itself may be unstable.
+
+### Tree structure
```
RaduNichita:
```suggestion
### Tree structure
```

Should we have a section more like "Developing"? For anybody only looking to use the tool most of the things beyond this point are not that useful.

tools/beman-tidy/README.md
neatudarius: Moved this section into `docs/dev-guide.md`.

tools/beman-tidy/README.md
neatudarius: Same would apply for check. First version will have 3/40 implemented. I would like to keep it, as my plan is to create lots of issues, and some will implement checks, some will implement fixes or both. Let's have the option exposed.

.github/workflows/beman-tidy.yml
neatudarius: "Tests" is OK. "Tests" is for testing the tool.

"Checks" is when you run the tool to check your code/repo.

Would help to say "Run Beman Tidy Unit Tests"? LE I don't think it will change something - Tests is for testing what you put in the repo - e.g. https://github.com/bemanproject/optional/blob/main/.github/workflows/ci.yml#L1

tools/beman-tidy/README.md
neatudarius: I think into a dev-guide.md doc we can have such details. I think it's OK to have a full description of the tree, especially since we'll have only 1-2 owners, and possible multiple ad-hoc contributors.

tools/beman-tidy/README.md
neatudarius: Solved by https://github.com/bemanproject/infra/pull/2#discussion_r2141078500?

tools/beman-tidy/images/usage-example.png
RaduNichita: Can you try using \Color if you need colourful output. I think GitHub supports this.

Otherwise, I am fine with keeping the image.

tools/beman-tidy/lib/checks/base/base_check.py
neatudarius: Check again.

tools/beman-tidy/images/usage-example.png
neatudarius: Removed images for now.

tools/beman-tidy/Makefile
neatudarius: I think your suggestion is great, but I propose to merge your PR after merging this one.

Created https://github.com/bemanproject/infra/issues/25. We'll be back for https://github.com/neatudarius/infra/pull/4, after this PR is merged. Is already big enough IMO.

@rishyak , do you agree? Priority should be to have v1 in main, and allow people to implement in parallel multiple checks, regardless of how we build /run the tool.

tools/beman-tidy/Makefile
neatudarius: Actually, it tries both tools from MAC OS and Linux:
```
	brew install autopep8 &> /dev/null
	apt-get install autopep8 -y &> /dev/null
	pip3 install autopep8
```

It seems that we can defer to a follow-up PR as descried here https://github.com/bemanproject/infra/pull/2#discussion_r2141150569. @rishyak , do you agree?

.github/workflows/beman-tidy.yml
RaduNichita: No, it's good how it is now. Resolving the comment.

tools/beman-tidy/README.md
RaduNichita: I think:

1) is hard to maintain, as everytime we add a new file we have to update the entire tree
2) it clutters the README with a bit of not relevant details (e.g - Makefile for the beman-tidy tool or Allows recursive Python packages imports)

I would prefer to keep the notes instead.

tools/beman-tidy/README.md
RaduNichita: replied.

tools/beman-tidy/README.md
neatudarius: Ok, removed

tools/beman-tidy/Makefile
steve-downey: While the intent, if I've followed correctly, is to run this in a container, please don't leave global install commands lying about. I need less splashover from the various projects I have, not more. Nothing should run `apt-get` or `brew` casually. `pip` can be run if you know you're in an active venv, but should take care that the venv is active before running.

tools/beman-tidy/Makefile
steve-downey: I am out of touch a bit with the python ecosystem, other than using a lot of tools available from installling in a venv. Do the current tools orchestrate setting up and pinning the environment reproducibly?  
I've heard nothing but good things about `uv` though!

tools/beman-tidy/README.md
steve-downey: Begin as you mean to proceed, in my opinion. If you want a --fix-inplace mode put it there from the start or something will make it more difficult later.

tools/beman-tidy/Makefile
rishyak: @steve-downey uv creates a [lockfile](https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile) `uv.lock` for reproducibility across machines. 

Excerpt from docs:
> `uv.lock` is a universal or cross-platform lockfile that captures the packages that would be installed across all possible Python markers such as operating system, architecture, and Python version.
>
> Unlike the `pyproject.toml`, which is used to specify the broad requirements of your project, the lockfile contains the exact resolved versions that are installed in the project environment. This file should be checked into version control, allowing for consistent and reproducible installations across machines.

More here: https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile

We can make it even more tool-anostic with `pylock.toml`(https://peps.python.org/pep-0751/).

You can even do cool things with uv and pre-commit:
https://docs.astral.sh/uv/guides/integration/pre-commit/

tools/beman-tidy/Makefile
rishyak: We will no longer need a Makefile if we migrate to PEP 518, which takes care of this problem all together. 
https://github.com/rishyak/infra/commit/3eae21c9fbbce2bbbdfdc2111242c5d83d55397e#diff-49d1de62ab0e3be12cde663db485e957ff16be46db9e8de5f926ab7341fbd36b 

Once we've switched to PEP 518, all of our build and dependency management is handled _declaratively_ (the big theme of this decade). After that, you're free to pick whichever tool you like- pip, Poetry, uv, Rye, or even your own local Makefile (just be sure it never lands in the repo or I, Rishyak, will be _very upset_).

Less code, less bug 😄

tools/beman-tidy/Makefile
nickelpro: Standardized lockfiles will be arriving to tooling Soon^TM thanks to [PEP 751](https://peps.python.org/pep-0751/).

`pip` is already [shipping experimental support](https://pip.pypa.io/en/stable/cli/pip_lock/) via `pip lock`. Once the rest of the frontends catch up there will be no tooling lock-in for lock files (say that 3 times fast). Users will be able to use whatever python project tool floats their particular boat.

rishyak: @neatudarius  - I'm flexible on whether PEP 518 support lands in its own PR, but I don't see any reason it can't be included here. Adopting it now gives us declarative, cross-platform builds out of the box, and means we never have to resort to `brew install` or `apt-get` when building. I'm with Steve on this.

tools/beman-tidy/.beman-standard.yml
JeffGarland: How is the .yml file maintained? It's just 'hand done'?  Seems like it could be scripted and maintained in bemat n -- then beman-module could be used to update here as discussed. Or something like that.  In the end it's a suggestion for improvement P1 fixes the issue I noticed...

neatudarius: Yes, it's hand done. 

It's a possibility to store it inside  the beman/ repo. Not sure it's the best idea, since we'll enforce people to update 2 files.

We may also change the format of the doc to be parser friendly- and have a single doc fetched by beman module. What do you think?

neatudarius: I will fix --fix-inplace as API from first release.

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+- id: beman-tidy
```
neatudarius, line 1: @wusatosi , integrated with pre-commit.

Here is a draft PR into exemplar using this tools, from my branch - https://github.com/bemanproject/exemplar/pull/181/files.

tools/beman-tidy/lib/checks/base/base_check.py
neatudarius: We'll do in https://github.com/bemanproject/infra/issues/30, not sure if it helps. I will still need this switch: I have 2 type of logs and I want my custom format.

tools/beman-tidy/lib/pipeline.py
neatudarius: Moved into https://github.com/bemanproject/infra/issues/30.

tools/beman-tidy/Makefile
neatudarius: OK, we'll switch to uv. Moved conversation to https://github.com/bemanproject/infra/pull/2#discussion_r2150059388

tools/beman-tidy/Makefile
neatudarius: Let's do the switch to uv.
- I will first merge this PR as it's already huge.
- We'll continue reviewing your, but relative to the updated main.

tools/beman-tidy/lib/checks/base/base_check.py
```diff
@@ -0,0 +1,102 @@
...
+        - Otherwise, it returns False and self.fix() must be able to fix the issue.
+        """
+        pass
+
+    @abstractmethod
+    def fix(self):
+        """
+        Fixes the issue if the Beman Standard is not applied.
+        - If check already applied, this method is a no-op and should return True.
+        - Otherwise, it will try to apply the check inplace. Returns the status of the fix attempt.
```
JeffGarland, line 94: Note that subclasses might not implement more than a stub if the fix method is too difficult to implement

JeffGarland (approved):

RaduNichita (approved):
LGTM

PR #5, opened 2025-03-14
Quick fix for `Compute Image Name`

wusatosi:
The original intend of the CI script is to upload devcontainer image to `bemanproject/devcontainers` instead of `bemanproject/infra`.

However there's some error in my code, for the details, see [this](https://github.com/bemanproject/infra/actions/runs/13866311408/job/38806052877#step:2:3) workflow log.

This PR fixes that.

### DIFF
--- a/.github/workflows/build_devcontainer.yml
+++ b/.github/workflows/build_devcontainer.yml
@@ -28,13 +28,13 @@ jobs:
             version: 14
           - kind: llvm
             version: 19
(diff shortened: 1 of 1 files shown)


neatudarius (approved):

wusatosi:
Why is the check not triggering?????

RaduNichita (approved):
The content seems ok to me, should we merge this and then investigate why the workflow is not triggered at all?

PR #9, opened 2025-04-29
Formatting and spelling cleanups to pass the exemplar lint check

ednolan:


wusatosi (approved):
Thanks for the work!

wusatosi:
Hum, why did the devcontainer related CI ran here.

wusatosi:
oh it's because it touches the devcontainer folder... okay this is working as intended

ednolan:
But (I believe) it pushed the images before the pull request was actually submitted. That should be fixed. It should build the images when the pull request is created, but it should only push to upstream once the request is actually approved.

wusatosi:
> But (I believe) it pushed the images before the pull request was actually submitted. That should be fixed. It should build the images when the pull request is created, but it should only push to upstream once the request is actually approved.

Yes, those images are pushed to `bemanproject/infra` instead of the "prod" `bemanproject/devcontainer`. In principle it's designed for pr reviewers to double check by pulling the container. Right now those containers are not really tagged with the pr number or the branch (just plain `gcc-15` so subsequent pr/ non main branch pushes overlaps the same tag), as we would need some kind of retention policy and cleanup script to avoid jamming GitHub if we properly identify which container come from which branch and pr.

ednolan:
Oh, okay. That system seems fine for now.

PR #11, opened 2025-04-29
Introduce test containers

wusatosi:
Should be good enough for a 1.0 roll out.

Yes there's a lot of stuff I should be doing that this PR does not implement.

1. move a lot of the logic and container building out of the CI so we can build images locally
2. support newer/ experimental versions of compilers, like gcc-15 or specific branch/ date release of llvm.
3. we should tag pre-release compilers with dates so it's more reproducible.
4. More built-in code coverage tool support.

But this should be sufficient work for us to do a 1.0 roll out. This covers most cases already.

POC:
- utf_view: https://github.com/bemanproject/utf_view/pull/44
- optional: https://github.com/bemanproject/optional/pull/121
- inplace_vector: https://github.com/bemanproject/inplace_vector/pull/83


ptsouchlos (approved):

.github/workflows/build_devcontainer.yml
```diff
@@ -50,11 +86,12 @@ jobs:
...
           build-args: |
-            compiler_kind=${{ matrix.kind }}
-            compiler_version=${{ matrix.version }}
+            base_image=${{ steps.image_base.outputs.image }}
+            compiler_kind=${{ matrix.compilers.kind }}
+            compiler_version=${{ matrix.compilers.version }}
           push: true
-          tags: ${{ env.REGISTRY }}/${{ steps.image_name.outputs.image_name }}:${{ matrix.kind }}-${{ matrix.version }}
+          tags: ${{ env.REGISTRY }}/${{ steps.image_name.outputs.image_name }}:${{ steps.image_name.outputs.tag }}
           # https://github.com/docker/build-push-action/issues/894
```
wusatosi, line 96: Let's move this all in a python script in the next iteration.

PR #12, opened 2025-05-05
Fix SPDX-License-Identifiers

ednolan:
"2.0 license with LLVM exceptions" seems to be a typo for "Apache-2.0 WITH LLVM-exception".


ptsouchlos (approved):
Based on this, it seems you are correct https://spdx.org/licenses/LLVM-exception.html

PR #16, opened 2025-05-27
Use update-alternatives

### LINKED ISSUE #8: Use update-alternatives
## Concern

`install_compiler.sh` manually manipulates files under `/usr/bin`, which typically is good to avoid:

```
sudo rm -f /usr/bin/gcc
sudo rm -f /usr/bin/g++

sudo ln -s "$(which gcc-"$VERSION")" /usr/bin/gcc
sudo ln -s "$(which g++-"$VERSION")" /usr/bin/g++
```

## Instead

Ubuntu provides a feature exactly for this use case called `update-alternatives`. We should consider using it.

The general idea is:

1. Install the desired compiler versions:
```
sudo apt update
sudo apt install gcc-VERSION g++-VERSION  # Replace VERSION with the desired version (e.g., 7, 8, 9)
```

2. Add the compilers as alternatives:

```
sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-VERSION 10  # Replace VERSION
sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-VERSION 10  # Replace VERSION
```

3. Configure the default compiler version:

```
sudo update-alternatives --config gcc
sudo update-alternatives --config g++
```

wusatosi:
Adopts suggestion from #8 .

Fixes #8

Edit: include a drive-by fix for triggering devcontainer build on PR.

### DIFF
--- a/.github/workflows/build_devcontainer.yml
+++ b/.github/workflows/build_devcontainer.yml
@@ -4,9 +4,14 @@ name: Publish Beman Containers
 
 on:
   push:
+    branches:
+      - "main"
     paths:
       - ".github/workflows/build_devcontainer.yml"
       - "containers/**"
+  pull_request:
+    branches:
+      - "main"
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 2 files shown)


ptsouchlos (approved):

ednolan:
Why hasn't CI run for this commit?

wusatosi:
> Why hasn't CI run for this commit?

Good question. Thanks for catching that. 

Edit: fixed, great, it caught errors!

wusatosi:
okay there's permission issues we need to fix with changes to docker containers from forks.
This is outside of the scope of this PR.
Consider the size of the PR is so small, I will just recreate this PR from a branch in bemanproject's infra repo.

Issue #14, opened 2025-05-13
container builder utility


wusatosi:
Right now most of the container building logic is in the CI file, we should move it out as a separate script so it's more testable and invokable locally.

PR #17, opened 2025-05-29
Use update-alternatives for installing compilers

### LINKED ISSUE #8: Use update-alternatives

wusatosi:
This is recreated from bemanproject/infra#16 due to permission complications with publishing containers to beman registry from a forked library.

Adopts suggestion from bemanproject/infra#8 .
Fixes bemanproject/infra#8 .

### DIFF
--- a/containers/install_compiler.sh
+++ b/containers/install_compiler.sh
@@ -10,16 +10,13 @@ echo "Install ${TOOL} at: ${VERSION}"
 
 shopt -s nocasematch
 if [ "$TOOL" = "gcc" ]; then
-    sudo apt-get remove -y gcc-"$VERSION" g++-"$VERSION" gcc g++
     sudo apt-get install -y gcc-"$VERSION" g++-"$VERSION" lcov
 
-    sudo rm -f /usr/bin/gcc
-    sudo rm -f /usr/bin/g++
-    sudo rm -f /usr/bin/gcov
+    sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-"$VERSION" 10
+    sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-"$VERSION" 10
 
-    sudo ln -s "$(which gcc-"$VERSION")" /usr/bin/gcc
-    sudo ln -s "$(which g++-"$VERSION")" /usr/bin/g++
-    sudo ln -s "$(which gcov-"$VERSION")" /usr/bin/gcov
+    sudo update-alternatives --config gcc
+    sudo update-alternatives --config g++
 
     gcc --version
 else
@@ -29,11 +26,11 @@ else
     sudo bash llvm.sh "${VERSION}"
     sudo apt-get install -y libc++-"$VERSION"-dev lcov
 
-    sudo rm -f /usr/bin/clang
-    sudo rm -f /usr/bin/clang++
+    sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-"$VERSION" 10
+    sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-"$VERSION" 10
 
-    sudo ln -s "$(which clang-"$VERSION")" /usr/bin/clang
-    sudo ln -s "$(which clang++-"$VERSION")" /usr/bin/clang++
+    sudo update-alternatives --config clang
+    sudo update-alternatives --config clang++
 
     clang --version
 fi


wusatosi:
CI fail should be external. Will retrigger once more CI events clear.

containers/install_compiler.sh
```diff
@@ -10,16 +10,13 @@ echo "Install ${TOOL} at: ${VERSION}"
...
-    sudo rm -f /usr/bin/g++
-    sudo rm -f /usr/bin/gcov
+    sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-"$VERSION" 10
+    sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-"$VERSION" 10
 
-    sudo ln -s "$(which gcc-"$VERSION")" /usr/bin/gcc
-    sudo ln -s "$(which g++-"$VERSION")" /usr/bin/g++
-    sudo ln -s "$(which gcov-"$VERSION")" /usr/bin/gcov
+    sudo update-alternatives --config gcc
+    sudo update-alternatives --config g++
```
wusatosi, line 19: Do we need this `config` step?

containers/install_compiler.sh
```diff
@@ -10,16 +10,13 @@ echo "Install ${TOOL} at: ${VERSION}"
...
 shopt -s nocasematch
 if [ "$TOOL" = "gcc" ]; then
-    sudo apt-get remove -y gcc-"$VERSION" g++-"$VERSION" gcc g++
     sudo apt-get install -y gcc-"$VERSION" g++-"$VERSION" lcov
 
-    sudo rm -f /usr/bin/gcc
-    sudo rm -f /usr/bin/g++
-    sudo rm -f /usr/bin/gcov
+    sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-"$VERSION" 10
+    sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-"$VERSION" 10
```
JeffGarland, line 16: This is one way -- but couldn't we just install each of the individual versions and then call them appropriately -- then 1 CI run with ubuntu could cover all the compilers?  Right now I have at least 4 compilers installed on my ubuntu -- g++14, g++15, clang-19 and clang++20.  There's actual no g++ or clang++ in /usr/bin.  

I'm completely fine with going this way, but just throwing this out there...

wusatosi: I tried that, that imagine is 7 GB (or 15 I forgot) and it took 1 minutes and a half to pull for GitHub actions.

JeffGarland: I see, so it's faster to do it in parallel with several runners -- got it.

JeffGarland (approved):

containers/install_compiler.sh
wusatosi: You're not the first one wanting this though, maybe I can try to implement this with #14

steve-downey: The gotcha is usually which libstdc++ and libc++ is in play at runtime.
For long term use, swapping out the system compiler will cause things to break down. Less of a concern for transient VMs.

containers/install_compiler.sh
steve-downey: That makes sure all of the other things associated with gcc and g++ are also configured to use the new alternative. If there are any.

containers/install_compiler.sh
```diff
@@ -29,11 +26,11 @@ else
     sudo bash llvm.sh "${VERSION}"
```
steve-downey, line 26: Consider `sudo bash llvm.sh "${VERSION}" all` 
so that all the related packages are installed also. 
This is probably why we don't find clang-scan-deps sometimes.

wusatosi: I am not aware of that! Thanks!

ednolan (approved):
I tested the images that infra CI built from the most recent commit on utf_view and it was able to build utf_view successfully.

JeffGarland:
This PR seems to be languishing here - @wusatosi are we ready to merge besides conflicts?

ednolan:
Superseded by https://github.com/bemanproject/infra-containers/pull/4

PR #20, opened 2025-06-02
grab clang tools from upstream llvm

nickelpro:
Does the thing we need for modules

### DIFF
--- a/containers/install_compiler.sh
+++ b/containers/install_compiler.sh
@@ -27,7 +27,7 @@ else
     wget https://apt.llvm.org/llvm.sh
 
     sudo bash llvm.sh "${VERSION}"
-    sudo apt-get install -y libc++-"$VERSION"-dev lcov
+    sudo apt-get install -y libc++-"$VERSION"-dev clang-tools-"$VERSION" lcov
 
     sudo rm -f /usr/bin/clang
     sudo rm -f /usr/bin/clang++
@@ -37,4 +37,3 @@ else
 
     clang --version
 fi
-


JeffGarland (approved):
LGTM

nickelpro:
The failures on this is because some of the CI machines rolled snake eyes and got an IP which is banned from pulling from upstream repos, nothing in this PR.

containers/install_compiler.sh
```diff
@@ -27,7 +27,7 @@ else
     wget https://apt.llvm.org/llvm.sh
 
     sudo bash llvm.sh "${VERSION}"
-    sudo apt-get install -y libc++-"$VERSION"-dev lcov
+    sudo apt-get install -y libc++-"$VERSION"-dev clang-tools-"$VERSION" lcov
```
wusatosi, line 30: I don't think this is right. the reason we are using `llvm.sh` is because we get to use llvm's upstream.

nickelpro: That's what this does. `llvm.sh` is adding the llvm upstream to apt, then we're pulling `libc++` and `clang-tools` from it.

Go look at how `llvm.sh` works, the final line is the one that installs llvm/clang

```
apt-get update
PKG="clang-$LLVM_VERSION lldb-$LLVM_VERSION lld-$LLVM_VERSION clangd-$LLVM_VERSION"
if [[ $ALL -eq 1 ]]; then
    # same as in test-install.sh
    # No worries if we have dups
    PKG="$PKG clang-tidy-$LLVM_VERSION clang-format-$LLVM_VERSION clang-tools-$LLVM_VERSION llvm-$LLVM_VERSION-dev lld-$LLVM_VERSION lldb-$LLVM_VERSION llvm-$LLVM_VERSION-tools libomp-$LLVM_VERSION-dev libc++-$LLVM_VERSION-dev libc++abi-$LLVM_VERSION-dev libclang-common-$LLVM_VERSION-dev libclang-$LLVM_VERSION-dev libclang-cpp$LLVM_VERSION-dev liblldb-$LLVM_VERSION-dev libunwind-$LLVM_VERSION-dev"
    if test $LLVM_VERSION -gt 14; then
        PKG="$PKG libclang-rt-$LLVM_VERSION-dev libpolly-$LLVM_VERSION-dev"
    fi
fi
apt-get install -y $PKG
```
We could set `ALL=1` and have the script install `clang-tools` for us, but that drags in other stuff we don't need.

PR #35, opened 2025-06-18
[beman-tidy] add check: README.LIBRARY_STATUS

neatudarius:
beman-tidy: add check - README.LIBRARY_STATUS - #48 

Note: This can be used an example of adding a file-based check (a check implementation which should validate a file from a repo given via a path).

Docs:
- [infra@tools/beman-tidy/docs/dev-guide.md - Adding a new check](https://github.com/bemanproject/infra/blob/main/tools/beman-tidy/docs/dev-guide.md#adding-a-new-check)
- [infra@tools/beman-tidy/docs/dev-guide.md - Testing](https://github.com/bemanproject/infra/blob/main/tools/beman-tidy/docs/dev-guide.md#testing)
- [infra@tools/beman-tidy/README.md - Public usage API](infra@tools/beman-tidy/README.md)


tools/beman-tidy/tests/beman_standard/readme/test_readme.py
```diff
@@ -83,3 +83,35 @@ def test__README_BADGES__invalid(repo_info, beman_standard_check_config, invalid
...
+    ]
+
+    file_testcases_run_valid(valid_readme_paths, ReadmeLibraryStatusCheck,
+                             repo_info, beman_standard_check_config)
+
+
+def test__README_LIBRARY_STATUS__invalid(repo_info, beman_standard_check_config):
+    """Test that an invalid README.md library status fails the check"""
+    invalid_readme_paths = [
+        Path("tests/beman_standard/readme/data/invalid/invalid.md"),
```
RaduNichita: Can you be more specific in path name to understand what is the difference between these cases?

neatudarius: I already did that. Thus, there is an extra file "invalid.md" with lots of errors

neatudarius: image

neatudarius: So we have specific invalid test case files (title wrong because : is missing, title wrong because the beman.library name is not used etc), and a generic one (where we put an error for each category - title, status line etc). 

Does it make sense?

PR #103, opened 2025-06-30
CODEOWNERS: remove @bemanproject/core-reviewers

ednolan:
core-reviewers is deprecated.


neatudarius (approved):

ptsouchlos:
Are we having individual codeowner files per repo instead?

ednolan:
That was/is the status quo. The CODEOWNERS file here is intended to be used for changes to the `infra` repo, not for other projects to link to for deduplicating their own CODEOWNERS file.

PR #114, opened 2025-07-06
[beman-tidy] Implement check DIRECTORY.SOURCES

neatudarius:
[beman-tidy] Implement check DIRECTORY.SOURCES - #66 

Notes:
- DIRECTORY.SOURCES was converted into a conditional REQUIREMENT - https://github.com/neatudarius/beman/blob/main/docs/BEMAN_STANDARD.md#directorysources
- We cannot implement a proper fix() method., instead we do a log.

Updates:
- Check implementation
- Tests


tools/beman-tidy/beman_tidy/.beman-standard.yml
```diff
@@ -111,30 +111,18 @@ CMAKE.AVOID_PASSTHROUGHS:
...
-    - directory_name: include
-    - regex: <put a regex here if possible - e.g. include/beman/${LIBRARY_NAME}>
 DIRECTORY.IMPLEMENTATION_HEADERS:
     - type: REQUIREMENT
-    - regex: <put a regex here if possible - e.g. include/beman/${LIBRARY_NAME}/detail or include/beman/${LIBRARY_NAME}/detail_/>
 DIRECTORY.SOURCES:
-    - type: REQUIREMENT
-    - directory_name: src
-    - regex: <put a regex here if possible - e.g. src/beman/${LIBRARY_NAME}>
+    - type: RECOMMENDATION # TODO: Why DIRECTORY.SOURCES is a RECOMMENDATION?
```
neatudarius:
```suggestion
    - type: REQUIREMENT
```

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
```diff
@@ -6,9 +6,19 @@
...
+# All checks in this file extend the DirectoryBaseCheck class.
+#
+# Note: DirectoryBaseCheck is not a registered check!
 class BemanTreeDirectoryCheck(DirectoryBaseCheck):
     """
-    Check if the directory tree is a Beman tree.
+    Check if the directory tree is a Beman tree: ${prefix_path}/beman/${short_name}.
+    Examples for a repo named "exemplar":
+    - include/beman/exemplar
+    - src/beman/exemplar
```
JeffGarland: This makes no accomadation for the case where the library is header only and doesn't have an src directory at all

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
```diff
@@ -6,9 +6,19 @@
...
+    - tests/beman/exemplar
+    - examples/
+    - docs/
+    - papers/
```
JeffGarland: similarly here -- if this is a repo that doesn't source the paper this directory will be omitted.

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
JeffGarland: hmm ok I see the extra logic below -- it's a little confusing how this is structured...

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
JeffGarland: should we have a way to signal beman tidy that the library is header only or that the paper directory isn't going to be present?  Like a .beman-tidy-overrides file? Not sure how the tool can tell the intent without something like that

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
neatudarius: The implementation already allow conditional. Should I remove these examples from docs?

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
neatudarius: We can add a config file in each library repo to opt in for some medata data
- header-only library
- papers directory etc

But would that extra complexity be useful for our developers?  To make it work, we need to make it a requirement.

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
JeffGarland: Sorry space cadet that this is just a comment -- maybe just say something about the optionality of these instead of just removing.

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
JeffGarland: > We can add a config file in each library repo to opt in for some medata data
> 
>     * header-only library
> 
>     * papers directory etc
> 
> 
> But would that extra complexity be useful for our developers? To make it work, we need to make it a requirement.

Trust me I don't want any more complication here, but I guess I don't understand how the tool can tell if the omission of src if 'an accident' or in the end an error on the users part.  And yes I see the hoops there to try and detect a 'possibly misnamed' src, but it's difficult to be comprehensive. And with papers if you emit a warning whenever it's missing that will be a thorn in the developer that isn't maintaining a paper (probably the majority of repos so far).

rishyak: I'm just thinking out loud here and haven't looked at the code in question bht:
Why can't we grab this information from the CMLs?

I'm sure the heuristics won't be perfect and perhaps it's too hard but it's worth an analysis.

neatudarius: CMakeLists files won't help us for papers/, docs/, etc.

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
neatudarius: done

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
neatudarius: In this week session, we decided we don't want to add a config for beman-tidy in each library repo, instead we'll do best effort for some checks.

I will explore the CMake parsing in a follow-up PR, if that's OK for you. CC: @JeffGarland 

I want to merge this PR and avoid conflicts with other DIRECTORY.* PRs.

JeffGarland: yep, all good

JeffGarland (approved):

PR #123, opened 2025-07-09
[beman-tidy] Add TOPLEVEL.CMAKE tests

vickgoodman:
Implementation for TOPLEVEL.CMAKE - #40 

- Added tests
- Changed path_runners.py to account for directory paths


tools/beman-tidy/tests/utils/path_runners.py
```diff
@@ -22,13 +22,24 @@ def run_check_for_each_path(
...
-        check_instance = check_class(repo_info, beman_standard_check_config)
-        check_instance.path = Path(path)
+        if os.path.isfile(path):
+            # For file checks, set the path directly on the check instance
+            check_instance = check_class(repo_info, beman_standard_check_config)
+            check_instance.path = Path(path)
+        else:
+            # For repo checks, modify the repo_info to point to the test directory
+            repo_info["top_level"] = Path(path)
+            check_instance = check_class(repo_info, beman_standard_check_config)
```
neatudarius: duplicated line


```
if os.path.isdir(path):
   repo_info["top_level"] = Path(path)
check_instance = check_class(repo_info, beman_standard_check_config)
if os.path.isfile(path):
     check_instance.path = Path(path)
 ```

File: tools/beman-tidy/tests/utils/path_runners.py [hunk 2]
```diff
@@ -22,13 +22,24 @@ def run_check_for_each_path(
...
+
         check_instance.log_level = True
 
-        assert check_instance.pre_check() is True, (
-            f"[{check_instance.__class__.__name__}] pre_check() failed for {path}"
-        )
+        if os.path.isfile(path):
```
neatudarius, line 37: why is this condition required? It should always.
maybe the log should be changed to 

```
assert check_instance.pre_check() is True, (
            f"[{check_instance.__class__.__name__}] pre_check() failed for {check_instance.path}"
        )
```

?


If yes, same for check(). please test

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/test_toplevel.py
```diff
@@ -0,0 +1,59 @@
...
+        valid_cmake_paths,
+        ToplevelCmakeCheck,
+        repo_info,
+        beman_standard_check_config,
+    )
+
+
+def test__TOPLEVEL_CMAKE__invalid(repo_info, beman_standard_check_config):
+    """
+    Test that repositories without CMakeLists.txt pass the check.
```
neatudarius:
```suggestion
    Test that repositories without CMakeLists.txt fail the check.
```

please use the same wording as in test_reamde.py

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/data/valid/complete_repo/README.md
neatudarius, line 1: in complete_repo/README.md it's enough to add a single line 

"Dummy content, we only tests that README is non.empty".

Same for LICENSE and CMakeLists.txt.

We don't need to have files in toplevel/data/ dir, because we only tests that 1) file exists (including name) and 2) is not empty.

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/data/valid/repo_with_cmake/CMakeLists.txt
neatudarius, line 1: I think we don't have any value by using `valid/repo_with_cmake` dir. It's already included in `complete_repo`. We only test for CmakeHere!


I would add these tests

1. data/valid/exemplar -> having all 3 files, each one with a line!
2. data/invalid/missing-CMAKE having only README + LICENSE, missing CMAKE -> it should fail TOPLEVEL.CMAKE
3. data/invalid/invalid-CMAKE having only README + LICENSE + empty CMAKE file -> it should fail TOPLEVEL.CMAKE
4. 3. data/invalid/invalid-CMAKE having only README + LICENSE + a valid CMAKE file but in a subdirectory -> it should fail TOPLEVEL.CMAKE

neatudarius (changes requested):

tools/beman-tidy/tests/utils/path_runners.py
vickgoodman: I tested but here is the issue

The condition is required because of the logic of TOPLEVEL checks
- ToplevelCmakeCheck.check() calls super().pre_check() (which is FileBaseCheck.pre_check())
- FileBaseCheck.pre_check() checks if the file exists at self.path
- For invalid test cases (like missing-CMAKE/ dir), there's no CMakeLists.txt file
- So pre_check() returns False because the file doesn't exist
- The test runner asserts pre_check() is True, which fails

The TOPLEVEL checks are file existence checks. For invalid cases:
- The whole point is that the file doesn't exist
- So pre_check() should return False for invalid cases
- But the test runner assumes pre_check() should always be True

So I believe the easy fix is to modify the assert in path_runners.py to skip pre_check() assertion for directory-based checks

What are your thoughts?

tools/beman-tidy/tests/utils/path_runners.py
vickgoodman: Done

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/test_toplevel.py
vickgoodman: Done

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/data/valid/complete_repo/README.md
vickgoodman: Got it

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/data/valid/repo_with_cmake/CMakeLists.txt
vickgoodman: Got it

vickgoodman:
It seems that the CMake linter(gersemi) is throwing an error for the empty CMake file for this test:
> 3. data/invalid/invalid-CMAKE having only README + LICENSE + empty CMAKE file -> it should fail TOPLEVEL.CMAKE

I believe there are a few solutions for example add a comment to the CMake file so it is not completely empty:
```
# Empty file used for testing
```

Or exclude the file from the ```.pre-commit-config.yaml``` file:
```
exclude: tests/lib/checks/beman_standard/toplevel/data/invalid/invalid-CMAKE1/CMakeLists.txt
```

Or fix linter behaviour to handle empty CMake files.

What do you think should be done? Thanks

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/data/invalid/invalid-CMAKE2/cmake/CMakeLists.txt
neatudarius, line 1: why this file does not have dummy content?

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/data/valid/exemplar/CMakeLists.txt
neatudarius, line 1: same , why not dummy content? TopLevel checks do not actually read file contents!

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/test_toplevel.py
```diff
@@ -0,0 +1,59 @@
...
+        valid_cmake_paths,
+        ToplevelCmakeCheck,
+        repo_info,
+        beman_standard_check_config,
+    )
+
+
+def test__TOPLEVEL_CMAKE__invalid(repo_info, beman_standard_check_config):
+    """
+    Test that repositories without CMakeLists.txt fail the check.
```
neatudarius:
```suggestion
    Test repositories with invalid CMakeLists.txt.
```

Not all of them are missing

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/test_toplevel.py
```diff
@@ -0,0 +1,59 @@
...
+)
+
+test_data_prefix = "tests/lib/checks/beman_standard/toplevel/data"
+valid_prefix = f"{test_data_prefix}/valid"
+invalid_prefix = f"{test_data_prefix}/invalid"
+
+
+def test__TOPLEVEL_CMAKE__valid(repo_info, beman_standard_check_config):
+    """
+    Test that repositories with CMakeLists.txt pass the check.
```
neatudarius:
```suggestion
    Test repositories with valid CMakeLists.txt.
```

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/data/invalid/invalid-CMAKE2/cmake/CMakeLists.txt
vickgoodman: Because it would fail the CMake linter

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/data/valid/exemplar/CMakeLists.txt
vickgoodman: Same, it would fail the CMake linter

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/test_toplevel.py
vickgoodman: Fixing

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/test_toplevel.py
vickgoodman: On it

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/data/invalid/invalid-CMAKE2/cmake/CMakeLists.txt
neatudarius: The linter can be disabled from onside the file

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/data/valid/exemplar/CMakeLists.txt
neatudarius: https://github.com/bemanproject/infra/pull/123#discussion_r2197323540

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/data/invalid/invalid-CMAKE2/cmake/CMakeLists.txt
neatudarius: Or put that note as a comment in the file

.pre-commit-config.yaml
```diff
@@ -16,9 +16,9 @@ repos:
   - repo: https://github.com/BlankSpruce/gersemi
     rev: 0.15.1
     hooks:
-    - id: gersemi
-      name: CMake linting
-
+      - id: gersemi
+        name: CMake linting
+        exclude: ^infra/tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/data/invalid/.*CMakeLists\.txt$
```
neatudarius: I don't think this is the right approach.

We should have a comment inside that CMakeLists to be ignore by the linter. Or a config in the tests/ directory to ignore some files. 

Please revert all changes in `.pre-commit-config.yaml`. This should be a generic config.

neatudarius (changes requested):

.pre-commit-config.yaml
vickgoodman: Thank you for the feedback! I do believe that keeping the pre-commit config generic is the better approach, but here is the issue:

The problematic file is an intentionally empty CMakeLists.txt in invalid-CMAKE1/ dir and is used to test the TOPLEVEL.CMAKE check. Empty files cause gersemi to fail with syntax errors.

I investigated several alternatives to avoid modifying the pre-commit config.

**Attempted Solutions:**
1. **Gersemi-specific config files** (`.gersemi.yaml`, `.gersemiignore`) - These don't exist; gersemi doesn't support ignore files
2. **Local pre-commit configs** - Not supported in subdirectories
3. **Directory-specific configs** - No mechanism available for gersemi

**Proposed Solutions:**
- Remove the empty CMake file test completely
- Keep the exclude rule in pre-commit config file
- Running pre-commit with the SKIP attribute for that specific test:
```
SKIP=gersemi pre-commit run --files tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/data/invalid/invalid-CMAKE1/CMakeLists.txt
```

I believe that gersemi, as any tool, has limitations and removing the test would remove important validation functionality.

What do you think the solution is?

.pre-commit-config.yaml
```diff
@@ -2,10 +2,10 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v5.0.0
     hooks:
-    -   id: trailing-whitespace
-    -   id: end-of-file-fixer
-    -   id: check-yaml
-    -   id: check-added-large-files
```
neatudarius: Please revert all change in this file! (no diff, not even whitespace!)

```
git checkout main -- .pre-commit-config.yaml
```

tools/beman-tidy/tests/lib/checks/beman_standard/toplevel/data/invalid/invalid-CMAKE1/.cmake-lint.yaml
```diff
@@ -0,0 +1 @@
+enabled: false
```
neatudarius: Cool, if it works, then LGTM

vickgoodman:
I believe it is ready for merging

neatudarius (approved):
LGTM! Ship it! 💯 

Thanks for hard work here!

.pre-commit-config.yaml
neatudarius: Last pushed solution with excluding *data/ directories is OK

PR #124, opened 2025-07-13
[beman-tidy] Implement check REPOSITORY.CODEOWNERS

neatudarius:
[beman-tidy] Implement check REPOSITORY.CODEOWNERS - #63 

We do sanity checks for .github/CODEOWNERS. We do a log for fix().

### DIFF
--- a/tools/beman-tidy/beman_tidy/.beman-standard.yml
+++ b/tools/beman-tidy/beman_tidy/.beman-standard.yml
@@ -30,7 +30,6 @@ REPOSITORY.NAME:
     - regex: <put a regex here if possible>
 REPOSITORY.CODEOWNERS:
     - type: REQUIREMENT
-    - default_group: "@bemanproject/core-reviewers"
 REPOSITORY.DISALLOW_GIT_SUBMODULES:
     - type: RECOMMENDATION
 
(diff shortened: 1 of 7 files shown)


tools/beman-tidy/beman_tidy/.beman-standard.yml
```diff
@@ -30,7 +30,6 @@ REPOSITORY.NAME:
     - regex: <put a regex here if possible>
 REPOSITORY.CODEOWNERS:
     - type: REQUIREMENT
-    - default_group: "@bemanproject/core-reviewers"
```
JeffGarland, line 33: @neatudarius I think we should be removing this now in repos?  I think I stumbled into one yesterday in cache_latest.

tools/beman-tidy/tests/lib/checks/beman_standard/readme/test_readme.py
```diff
@@ -125,9 +125,6 @@ def test__README_BADGES__invalid(repo_info, beman_standard_check_config):
 
 @pytest.mark.skip(reason="NOT implemented")
 def test__README_BADGES__fix_inplace(repo_info, beman_standard_check_config):
-    """
```
JeffGarland, line 128: assuming the removal because these are tests we don't want these in generated docs, yes?

JeffGarland:
This LGTM -- ship it.

JeffGarland (approved):

rishyak (approved):

tools/beman-tidy/tests/lib/checks/beman_standard/readme/test_readme.py
neatudarius: Yes. We want to track these tests as not implemented, but we don't need to document them.

tools/beman-tidy/beman_tidy/.beman-standard.yml
neatudarius: The group is already removed from all repos. Can you point to a left over? I can also clean that.

PR #134, opened 2025-07-15
[beman-tidy] Implement DIRECTORY.INTERFACE_HEADERS check

vickgoodman:
Implement DIRECTORY.INTERFACE_HEADERS check - #43 

- Added check
- Added tests


tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
```diff
@@ -19,7 +20,55 @@ def __init__(self, repo_info, beman_standard_check_config, prefix_path):
...
+
+    def check(self):
+        """
+        Check that all public header files reside within the include/beman/<short_name>/ directory.
+        """
+        # Check if the path exists.
+        # Example path: "exemplar/include/beman/exemplar"
+        include_path = self.path / "include" / "beman" / self.repo_info["name"]
+        if (
+            not os.path.exists(include_path)
```
neatudarius:
```suggestion
            not include_path.exists()
```

include_path is already a Path. We can remove importing os.

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
```diff
@@ -19,7 +20,55 @@ def __init__(self, repo_info, beman_standard_check_config, prefix_path):
...
+        exclude_dirs = {"src"}
+        if self.repo_info["name"] == "exemplar":
+            exclude_dirs.add("cookiecutter")
+
+        hpp_files = []
+        for root, dirs, files in os.walk(self.path):
+            dirs[:] = [d for d in dirs if d not in exclude_dirs]
+
+            for name in files:
+                if name.lower().endswith(".hpp"):
```
neatudarius: We should also search for `.hpp`, `h`, `.hh`, `.hxx`.

tools/beman-tidy/tests/lib/checks/beman_standard/directory/data/valid/repo-exemplar-v1/include/beman/repo-exemplar-v1/identity.hpp
neatudarius, line 1: Please add `// Dummy content, this file is used only for testing.` in all header files.

tools/beman-tidy/tests/utils/path_runners.py
```diff
@@ -52,6 +52,8 @@ def run_check_for_each_path(
         if os.path.isdir(path):
             # For repo checks, modify the repo_info to point to the test directory
             repo_info["top_level"] = Path(path)
+            # Set the last part of the path as the repository name
+            repo_info["name"] = path.parts[-1]
```
neatudarius: why do we need to do that ? should you use run_check_for_each_repo instead? or create Other function. But this change is not OK in `run_check_for_each_path`

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
```diff
@@ -19,7 +20,55 @@ def __init__(self, repo_info, beman_standard_check_config, prefix_path):
         )
 
 
-# TODO DIRECTORY.INTERFACE_HEADERS
+@register_beman_standard_check("DIRECTORY.INTERFACE_HEADERS")
+class DirectoryInterfaceHeadersCheck(DirectoryBaseCheck):
```
neatudarius:
```suggestion
class DirectoryInterfaceHeadersCheck(BemanTreeDirectoryCheck):
```

Should use this helper class - check https://github.com/bemanproject/infra/pull/114

tools/beman-tidy/tests/lib/checks/beman_standard/directory/data/invalid/repo-exemplar-v3/include/beman/repo-exemplar-v3/identity.hpp
neatudarius, line 1: Something is wrong with entries in `data/invalid`.

Btw, I already added in https://github.com/bemanproject/infra/pull/134, 3 versions of invalid repos. I thing that PR may get in first, sorry.

neatudarius (changes requested):
Let's sync about this PR, I have some concerns

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
vickgoodman: I wanted to use this helper class but it sets working directory to this:
`f"{prefix_path}/beman/{repo_info['name']}",`
but I need to do the check in the whole repo, not just in exemplar/include/beman/examplar to use as an example.

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
neatudarius: still not applied - we should not require using `os` if we are using `Path`

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
neatudarius: You should not have `include_path = self.path / "include" / "beman" / self.repo_info["name"]` in your code!
This is why `BemanTreeDirectoryCheck` class exists. We should have `class DirectoryInterfaceHeadersCheck(BemanTreeDirectoryCheck)`, which will give you `self.path` -> `include/beman/short_name`. And then you can create other paths, if you want starting from `self.repo_path`.

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
```diff
@@ -30,7 +31,55 @@ def __init__(self, repo_info, beman_standard_check_config, prefix_path):
...
+    def check(self):
+        """
+        Check that all public header files reside within the include/beman/<short_name>/ directory.
+        """
+        # Check if the path exists.
+        # Example path: "exemplar/include/beman/exemplar"
+        include_path = self.path / "include" / "beman" / self.repo_info["name"]
+        if (
+            not os.path.exists(include_path)
+            or os.path.isfile(include_path)
```
neatudarius:
```suggestion
            or os.path.isfile(include_path)
```
can we use pathlib instead os everywhere?

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
```diff
@@ -30,7 +31,55 @@ def __init__(self, repo_info, beman_standard_check_config, prefix_path):
...
+            or len(os.listdir(include_path)) == 0
+        ):
+            self.log(
+                f"The path '{self.path}' does not exist, is a file or is empty."
```
neatudarius:
```suggestion
                f"The path '{self.path}' does not exist, is a file or is empty. "
```

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
```diff
@@ -30,7 +31,55 @@ def __init__(self, repo_info, beman_standard_check_config, prefix_path):
...
+                " All public header files must reside within include/beman/<short_name>/."
```
neatudarius:
```suggestion
                "All public header files must reside within include/beman/<short_name>/."
```

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
```diff
@@ -30,7 +31,55 @@ def __init__(self, repo_info, beman_standard_check_config, prefix_path):
...
+            )
+            return False
+
+        # Get all .hpp files paths, excluding certain directories.
+        exclude_dirs = {"src"}
+        if self.repo_info["name"] == "exemplar":
+            exclude_dirs.add("cookiecutter")
+
+        hpp_files = []
```
neatudarius:
```suggestion
        header_files = []
```

Please don't resolve comments , just push fixes and reply. I will marke comments as resolved e.g. https://github.com/bemanproject/infra/pull/134/files#r2207970309.

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
```diff
@@ -30,7 +31,55 @@ def __init__(self, repo_info, beman_standard_check_config, prefix_path):
...
+
+            for name in files:
+                if name.lower().endswith(".hpp"):
+                    hpp_files.append(os.path.join(root, name))
+
+        # Check that all .hpp files are under the include/beman/<short_name>/
+        for hpp_file in hpp_files:
+            if not hpp_file.startswith(str(include_path)):
+                self.log(
+                    f"Header file '{hpp_file}' is not under include/beman/{self.repo_info['name']}/ directory."
```
neatudarius:
```suggestion
                    f"Header file '{hpp_file}' is not under include/beman/{self.repo_name}/ directory."
```
we have that shortcut - pleas read `tools/beman-tidy/beman_tidy/lib/checks/base/base_check.py`

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
```diff
@@ -30,7 +30,56 @@ def __init__(self, repo_info, beman_standard_check_config, prefix_path):
...
+        super().__init__(repo_info, beman_standard_check_config, "include")
+
+    def pre_check(self):
+        # Need to override this, because DIRECTORY.INTERFACE_HEADERS is conditional
+        # (a repo without any interface header files is still valid).
+        return True
+
+    def check(self):
+        """
+        Check that all public header files reside within the include/beman/<short_name>/ directory.
```
neatudarius, line 45: Again, let's not discussed implementation from this PR, as the pattern is similar to the one for PAPERS (https://github.com/bemanproject/infra/pull/159#discussion_r2223086644).

I propose a plan for DIRECTORY.INTERFACE_HEADERS and DIRECTORY.IMPLEMENTATION_HEADERS
image

@valid header locations
First of all, we can have headers in
1. include/beman/<short_name> - valid location
2. include/ but not in include/beman/<short_name> - invalid location
3. src/beman/short_name - valid location
4. src but not in src/beman/short_name  - invalid location
5. examples/ - valid location
6. tests/beman/short_name - valid location
7. tests/ but not in tests/beman/short_name - invalid location
8. cookiecutter - valid location
9. infra - valid location
TLDR:
```Python
valid_header_locations = [
# Library specific
"include/beman/short_name",
"src/beman/short_name",
"tests/beman/short_name",
'examples",
# Infra
"infra",
"cookiecutter",
]
```
We can assume, IMO, for now, that this is a complete set of locations which can be used in v1 implementation. If we go with this approach, we can actually run locally run the tool on all libraries and see if we are missing something before first commit!

@CML can provide us necessary info?
1) I briefly investigated if we can use CML. As per my understanding, CML does not export this info in compile_commands.json, so I think we should actually parse the CML and find the `target_sources` call and see what we set for `PUBLIC` and `PRIVATE`. @nickelpro ?
2) I don't want to complicate things - let's assume it works with CML. What if a library requires custom build flags to actually build with a (sub)directory?  etc
I think it's better to do best-effort search by accessing the filesystem, IMO.

=================================================================
My proposal

I. `Steps for DIRECTORY.INTERFACE_HEADERS`:
1. We assert to exists `include/beman/<short_name>` (we cannot have a library without at least one public header!). If invalid, we stop -> `Missing include/beman/<short_name>. Expected public interface to be present. Please check ...`.
2. We do search for all headers in the repo which are not in `valid_header_locations` , and if we find one we fail the check.
3. If 1 + 2 passed, `DIRECTORY.INTERFACE_HEADERS` check also passes. 
(We actually just checked headers are at right location. We didn't use the "public" thing here. It's not needed IMO.)
TLDR: It's same proposal from this PR.

II. `Steps for DIRECTORY.IMPLEMENTATION_HEADERS`. 
We cannot actually easier distinguish which is public and which is private headers. We can say it's SKIPPED and log a message (we assume my PR #148 is already merged).

=================================================================

Again, please ignore the current pushed implementation, let's decide what should be the desired approach.
CC: @vickgoodman (PR owner) @JeffGarland (Style Czar) @nickelpro (Build Czar)

neatudarius (changes requested):
temporary blocking this PR until we figured out what's desired plan and expectations for tidy and headers checks

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
neatudarius: Due to Victor's availability (PTO), we discussed and I may take this PR (we want DIRECTORY.*) to be completed for release 1.0 ASAP. @vickgoodman , can you confirm?

vickgoodman: Feel free to update the PR and close the issue. Thanks!

neatudarius:
Reminder to update this PR with latest main after https://github.com/bemanproject/infra/pull/170.

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
neatudarius: I will update this PR next weeks.

neatudarius:
> Reminder to update this PR with latest main after https://github.com/bemanproject/infra/pull/170.

@vickgoodman , reminder^ - the repo was renamed, the main history was changed.

Issue #81, opened 2025-06-18
[beman-tidy] Implement release.godbolt_trunk_version


neatudarius:
Note: We may not be able to (easily) integrate this checks in the tools (internet access and complexity).

release.godbolt_trunk_version from [beman_standard.md](https://github.com/bemanproject/beman/blob/main/docs/beman_standard.md#releasegodbolt_trunk_version)
- [x] Implement check() (dry-run mode)
- [x] Implement fix (fix-inplace mode)

Docs:
- [infra@tools/beman-tidy/docs/dev-guide.md - Adding a new check](https://github.com/bemanproject/infra/blob/main/tools/beman-tidy/docs/dev-guide.md#adding-a-new-check)
- [infra@tools/beman-tidy/docs/dev-guide.md - Testing](https://github.com/bemanproject/infra/blob/main/tools/beman-tidy/docs/dev-guide.md#testing)
- [infra@tools/beman-tidy/README.md - Public usage API](https://github.com/bemanproject/infra/blob/main/tools/beman-tidy/README.md#beman-tidy-the-codebase-bemanification-tool)

neatudarius:
Offline tool, we'll do dummy implementation in v1. TBD for v2- #146

PR #143, opened 2025-07-19
[beman-tidy] Implement RELEASE.*

neatudarius:
#81  #47 #82 

[beman-tidy] Implement RELEASE.*, or better say don't implement them as we cannot.
We still need to integrate them into beman-tidy, to at least log.
image

### DIFF
--- a/tools/beman-tidy/tests/lib/checks/beman_standard/release/test_release.py
+++ b/tools/beman-tidy/tests/lib/checks/beman_standard/release/test_release.py
@@ -12,6 +12,7 @@ from tests.utils.path_runners import (
 from beman_tidy.lib.checks.beman_standard.release import (
     ReleaseGithubCheck,
     ReleaseGodboltTrunkVersionCheck,
+    ReleaseNotesCheck,
 )
 
 test_data_prefix = "tests/lib/checks/beman_standard/release/data"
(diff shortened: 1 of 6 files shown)


neatudarius:
@JeffGarland , can you review this dummy implementation for RELEASE.*? Basically we already agreed in one meeting, that some checks cannot be implemented. We actually gave RELEASE.* as an example.

tools/beman-tidy/tests/lib/checks/beman_standard/release/test_release.py
```diff
@@ -0,0 +1,140 @@
...
+        beman_standard_check_config,
+    )
+
+
+@pytest.mark.skip(reason="NOT implemented")
+def test__RELEASE_GITHUB__invalid(repo_info, beman_standard_check_config):
+    """
+    Test that repositories with missing GitHub release fail the check.
+    """
+    # RELEASE.GITHUB always passes, as beman-tidy is an offline tool.
```
JeffGarland, line 47: should we ultimately have a version that can be optionally be online and check the release? it's probably not worth it

JeffGarland (approved):
looks fine

tools/beman-tidy/tests/lib/checks/beman_standard/release/test_release.py
neatudarius: Created an issue for beman-tidy v2. https://github.com/bemanproject/infra/issues/146#issuecomment-3092633777

Issue #144, opened 2025-07-19
[beman-tidy] LICENSE.APACHE_LLVM check is broken on main


ednolan:
```
~/sync/cpp/bemanproject/infra/tools/beman-tidy →main $ source ~/stuff/2025-07/beman-tidy-venv-20250714/bin/activate
~/sync/cpp/bemanproject/infra/tools/beman-tidy →main $ uv build
Building source distribution...
Building wheel from source distribution...
Successfully built dist/beman_tidy-0.1.0.tar.gz
Successfully built dist/beman_tidy-0.1.0-py3-none-any.whl
~/sync/cpp/bemanproject/infra/tools/beman-tidy →main $ python3 -m pip install dist/beman_tidy-0.1.0-py3-none-any.whl
// ...
~/sync/cpp/bemanproject/exemplar →main $ beman-tidy --no-fix-inplace --verbose --no-require-all .
beman-tidy pipeline started ...

Running check [REQUIREMENT][LICENSE.APPROVED] ...
        check [REQUIREMENT][LICENSE.APPROVED] ... PASSED

Running check [RECOMMENDATION][LICENSE.APACHE_LLVM] ...
Traceback (most recent call last):
  File "/home/eddie/stuff/2025-07/beman-tidy-venv-20250714/bin/beman-tidy", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/eddie/stuff/2025-07/beman-tidy-venv-20250714/lib/python3.13/site-packages/beman_tidy/cli.py", line 64, in main
    failed_checks = run_checks_pipeline(
        checks_to_run, args, beman_standard_check_config
    )
  File "/home/eddie/stuff/2025-07/beman-tidy-venv-20250714/lib/python3.13/site-packages/beman_tidy/lib/pipeline.py", line 139, in run_checks_pipeline
    ) = run_pipeline_helper()
        ~~~~~~~~~~~~~~~~~~~^^
  File "/home/eddie/stuff/2025-07/beman-tidy-venv-20250714/lib/python3.13/site-packages/beman_tidy/lib/pipeline.py", line 97, in run_pipeline_helper
    check_type, passed = run_check(implemented_checks[check_name])
                         ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #139, opened 2025-07-18
[beman-tidy] Implement LICENSE.APACHE_LLVM

vickgoodman:
As discussed in https://github.com/bemanproject/infra/pull/135#discussion_r2210507761

Implemented LICENSE.APACHE_LLVM - #55 

- Added check
- Added fix(just a log message)
- Added tests

PR #145, opened 2025-07-19
[beman-tidy] beman-tidy should access configuration files that are only stored inside the beman_tidy/ subdirectory

neatudarius:
[beman-tidy] beman-tidy should access configuration files that are only stored inside the beman_tidy/ subdirectory - #144 

#55 introduced the bug in #139


ednolan (approved):

neatudarius:
FYI @vickgoodman , read this fix!

Issue #147, opened 2025-07-20
[beman-tidy] Improve logging with SKIPPED status for dummy implementations


neatudarius:
[beman-tidy] Improve logging with SKIPPED status for dummy implementations

Right now:

<img width="1727" height="77" alt="Image" src="https://github.com/user-attachments/assets/8b54f25b-88b9-42b9-a749-36ca24210930" />
We should say that check is SKIPPED instead.

PR #148, opened 2025-07-20
[beman-tidy] Add SKIPPED as a possible status for a check

neatudarius:
#147 beman-tidy] Add SKIPPED as a possible status for a check

Updates:

1. Proposed a pattern to skip a chec: `BaseCheck.should_skip(): return False` method which can be overridden in derived classes: https://github.com/bemanproject/infra/pull/148/files#diff-3bdc17c1450533d17e01fa9a7048a1088909117a964cc9f0986db4701bf7ca50R85-R91. If this method returns true, `check()` and `fix()` are not called , but should exists. `pre_check()` is also not called.

2. Applied this proposal to LIBRARY.CRITERIA, which should be SKIPPED.

4.  Change the logging / output format for the tool to use SKIPPED category. Coverage formulas are also changed.

5. Update `docs/dev-guide.md` to specify when should write proper tests vs skipped tests. We always need a set of tests to know the explicit state of the tool.

6. Add tests to check your tests! Because in Python everything is an object, including the types, I added tests in `tests/lib/checks/system/test_registry.py`, which essentially checks that for __ALL__ registered check classes the proper set of tests is defined at the right location! a.k.a. we won't have anymore situations like in the past, were some checks were added in Sofia without tests.
Examples of failures:
- If I delete `test__LICENSE_CRITERIA__skipped` from this branch (or move it to wrong location)
image
- Or if I have wrong function name:
image

New logging format examples:
- non-verbose, exemplar, failing 1 x RECOMMANDATION
image
- non-verbose, exemplar with --require-all, failing 1 x REQUIREMENT (simplified output)
image

- verbose, exemplar, failing 1 x RECOMMANDATION - check `SKIPPED` new tag
image
- verbose, exemplar with --require-all, failing 1 x REQUIREMENT (simplified output) - check that `WARNING` is converted to `ERROR`
image

### DIFF
--- a/tools/beman-tidy/beman_tidy/lib/checks/base/base_check.py
+++ b/tools/beman-tidy/beman_tidy/lib/checks/base/base_check.py
@@ -54,9 +54,15 @@ class BaseCheck(ABC):
         )
         assert self.full_text_body is not None
 
-        # set log level - e.g. "ERROR" or "WARNING"
-        self.log_level = "ERROR" if self.type == "REQUIREMENT" else "WARNING"
+        # set log level - e.g. "ERROR" or "WARNING" or "SKIPPED"
         self.log_enabled = False
+        self.log_level = (
+            "SKIPPED"
+            if self.should_skip()
+            else "ERROR"
+            if self.type == "REQUIREMENT"
+            else "WARNING"
+        )
 
         # set repo info
         self.repo_info = repo_info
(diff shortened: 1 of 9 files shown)


neatudarius:
@JeffGarland , please take a look at screenshots from the PR description or give a try with the tool. I think the new format is way better, thanks for the SKIPPED suggestion!

RaduNichita (approved):

neatudarius:
I'm merging this PR as it was already reviewed and we don't have ongoing discussions here. I will do a demo during Weekly Beman Sync and if needed, apply feedback.

Issue #70, opened 2025-06-18
[beman-tidy] Implement directory.papers


neatudarius:
directory.papers from [beman_standard.md](https://github.com/bemanproject/beman/blob/main/docs/beman_standard.md#directorypapers)
- [x] Implement check() (dry-run mode)
- [x] Implement fix (fix-inplace mode)

Docs:
- [infra@tools/beman-tidy/docs/dev-guide.md - Adding a new check](https://github.com/bemanproject/infra/blob/main/tools/beman-tidy/docs/dev-guide.md#adding-a-new-check)
- [infra@tools/beman-tidy/docs/dev-guide.md - Testing](https://github.com/bemanproject/infra/blob/main/tools/beman-tidy/docs/dev-guide.md#testing)
- [infra@tools/beman-tidy/README.md - Public usage API](https://github.com/bemanproject/infra/blob/main/tools/beman-tidy/README.md#beman-tidy-the-codebase-bemanification-tool)

neatudarius:
<img width="1085" height="302" alt="Image" src="https://github.com/user-attachments/assets/2d85722f-25df-471d-82ec-a676be0d54ea" />
I think we can only do heuristic, because they can try to write papers with any technology...
H1:  If submodule wg21 is used, then it should be under `papers/` (e.g., papers/path/to/wg21). e.g., https://github.com/bemanproject/optional/tree/main/papers/P2988.
H2: All `*.tex` / `*.bib` files should be inside `papers/`. e.g., https://github.com/bemanproject/optional/tree/main/papers/P2988

FYI @vickgoodman

neatudarius:
@vickgoodman , please also add `*.md` in https://github.com/bemanproject/infra/blob/main/tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py#L217. (ofc, we need to exclude root README.md) and things in `docs/*` and `examples/*`.

PR #159, opened 2025-07-22
[beman-tidy] Implement DIRECTORY.PAPERS

vickgoodman:
Implemented DIRECTORY.PAPERS - #70 

- Added check
- Added fix(log messages)
- Added tests


tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
```diff
@@ -175,7 +172,16 @@ def check(self):
             exclude_dirs.append("cookiecutter")
 
         # File extensions that are considered "paper-related"
-        paper_extensions = [".bib", ".pdf", ".tex", ".png", ".jpg", ".jpeg", ".svg", ".bst"]
+        paper_extensions = [
+            ".bib",
```
JeffGarland, line 218: missing at least docx, org, html, css.  I'm questioning if this is a good idea.  Why do we need to try and enforce what is here?  For example, if someone was using asciidoc they might have .asc, .asciidoc, .ad, or .ascdoc.

vickgoodman: Well, my first thought for the check was to do a regex on paper numbers(eg. P2988/), but then I figured paper names can be different(eg. wg21), so I landed on this extensive file extension check. The check could be simplified to check just for paper numbers, but then it wouldn't be complete.

vickgoodman: So maybe a combination of regex and file extension search could be implemented? What do you think?

ednolan: What about the following check?

- There is a directory called 'papers'
- It contains a file whose filename contains at least four consecutive digits ("at least" to avoid the Y2K problem when we eventually exceed P9999)

Yes, that check could be considered "incomplete," but I don't think it matters-- we should assume that beman implementers are acting in good faith here, and just try to catch simple mistakes, not perform exhaustive checks that lead to overly prescriptive requirements.

vickgoodman: Ok, got it!

neatudarius: I think we should discuss again docs/ + papers/ (DIRECTORY.DOCS also can be improved in term of regex match here https://github.com/bemanproject/infra/blob/main/tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py#L119).
image


I don't see the thing as a requirement, but as a best effort implementation - which we actually agreed in the Beman Sync! (we agreed for a best effort implementation, not for a dumb one). 

Let me explain the intention here with an example - for instance, let's say `*md` is in our list `extensions` list for docs and papers, and `*tex` is not.
- If we'll found `*.md` files, we require them to only be in `docs/` and `papers/` (`DIRECTORY.DOCS` + `DIRECTORY.PAPERS`).
- We won't  find `*tex` at all. So we don't say anything! (it's a gap in the tool which will be reported and implemented).
Ofc: ROOT README and maybe others are exceptions.


P1. I think we should have meaningful checks and feedback from the tool. So I would go with the current proposed direction for multiple checks. 
P2. We can agree on a list of doc extensions to follow. If it's not inclusive enough, it won't restrict the users! It's a n omission from beman-tidy checks, which can be solved later.
docs_extensions = [
            ".md",
            ".bib",
            ".bst",
            ".tex",
            ".pdf"
            ".docs",
             ".docs",
             ".org",
              ".html",
              ".css",
               ...
]


Let's focus on what we want and please ignore the implementation for the moment. I would really want to implement meaningful feedback in this tool. CC: @JeffGarland @ednolan @vickgoodman

neatudarius (changes requested):
Temporary blocking this PR until we finish controversial thread.

tools/beman-tidy/beman_tidy/lib/checks/beman_standard/directory.py
ednolan: Oh okay, I retract my comments entirely in that case-- I misunderstood the polarity of what this check was doing. I thought it would fail the check if it *didn't* find the extensions *inside* the papers directory, not if it *did* find the extensions *outside* the papers directory.

neatudarius: > Oh okay, I retract my comments entirely in that case-- I misunderstood the polarity of what this check was doing. I thought it would fail the check if it didn't find the extensions inside the papers directory, not if it did find the extensions outside the papers directory.

Better comments we can do for sure.

neatudarius: I assume @JeffGarland's confusion was similar (as he said `enforce what is here`, which we don't do!).

We'll wait one more day, and if we don't have any other concerns, we'll moved with proposal from https://github.com/bemanproject/infra/pull/159#discussion_r2223086644.

CC: @vickgoodman

neatudarius: We did wait 2 days, as a lead for beman-tidy, I decide to go forward with [this](https://github.com/bemanproject/infra/pull/159#discussion_r2223086644) plan. @vickgoodman , please do the updates and I will review myself the PR. Thanks everyone!

neatudarius (approved):

PR #171, opened 2025-08-11
Add a beman_install_library CMake function

bretbrownjr:
Problem
-------

It requires to much code with too many accurate details to install a Beman library. Also, repeating those lines of code across many libraries makes it too difficult to maintain installation logic across the project.

Solution
--------

Create beman_install_library, a CMake function for installing a Beman library. This function uses Beman Standard assumptions to make it simple for Beman libraries to mark themselves for installation.

### DIFF
--- a/cmake/beman-install-library-config.cmake
+++ b/cmake/beman-install-library-config.cmake
@@ -1,0 +1,147 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+include_guard(GLOBAL)
+
+# This file defines the function `beman_install_library` which is used to
+# install a library target and its headers, along with optional CMake
+# configuration files.
+#
+# The function is designed to be reusable across different Beman libraries.
+
+function(beman_install_library name)
+    # Usage
+    # -----
+    #
+    #     beman_install_library(NAME)
+    #
+    # Brief
+    # -----
+    #
+    # This function installs the specified library target and its headers.
+    # It also handles the installation of the CMake configuration files if needed.
+    #
+    # CMake variables
+    # ---------------
+    #
+    # Note that configuration of the installation is generally controlled by CMake
+    # cache variables so that they can be controlled by the user or tool running the
+    # `cmake` command. Neither `CMakeLists.txt` nor `*.cmake` files should set these
+    # variables directly.
+    #
+    # - BEMAN_INSTALL_CONFIG_FILE_PACKAGES:
+    #      List of packages that require config file installation.
+    #      If the package name is in this list, it will install the config file.
+    #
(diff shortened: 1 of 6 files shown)


cmake/beman-install-library-config.cmake
neatudarius, line 1: I had the impression I already wrote this comment: can we make a `cmake/beman` or `cmake/infra` subdirectory? and a `cmake/toolchain` one?

neatudarius (approved):

ednolan (approved):

cmake/beman-install-library-config.cmake
neatudarius: @ednolan , any opinion here? 

e.g.
`cmake/infra/install-library-config.cmake`/`cmake/infra/use-fetch-content.cmake`
and 

`cmake/toolchain/gnu.cmake`/`cmake/toolchain/llvm.cmake`/

ednolan: That seems reasonable to me but we should do it as a separate follow-on commit to this one. It will require updates to CI and README files.

cmake/beman-install-library-config.cmake
```diff
@@ -0,0 +1,147 @@
...
+    # ---------------
+    #
+    # Note that configuration of the installation is generally controlled by CMake
+    # cache variables so that they can be controlled by the user or tool running the
+    # `cmake` command. Neither `CMakeLists.txt` nor `*.cmake` files should set these
+    # variables directly.
+    #
+    # - BEMAN_INSTALL_CONFIG_FILE_PACKAGES:
+    #      List of packages that require config file installation.
+    #      If the package name is in this list, it will install the config file.
```
JeffGarland, line 32: could we have an example -- I can read and understand all the english but I don't know what this does. maybe an example?

cmake/beman-install-library-config.cmake
```diff
@@ -0,0 +1,147 @@
...
+        message(
+            DEBUG
+            "beman-install-library: Installing a config package for '${name}'"
+        )
+
+        include(CMakePackageConfigHelpers)
+
+        find_file(
+            config_file_template
+            NAMES "${package_name}-config.cmake.in"
```
JeffGarland, line 101: oh, this dreaded thing. I won't hold this review up for this debate obviously, but I believe we should listen to this man on the subject.  Which is to say that any config input for the library should be a .hpp that is potentially generated or created by the user.

https://vector-of-bool.github.io/2020/10/04/lib-configuration.html

JeffGarland (approved):

bretbrownjr:
So the relevant `beman.exemplar` PR checked out great:
https://github.com/bemanproject/exemplar/pull/241

The latest commit is just a squashed and rebased commit of the already-reviewed contents of this PR.

PR #190, opened 2025-09-07
Polish beman_install_library

bretbrownjr:
This commit:

* Changes defaults such that one must opt out of installing CMake config packages for beman libraries using
beman_install_library. This is because if one is installing a beman library, it is more likely than not that they would want to support potential CMake consumers using those libraries.

* Fixes a bug that caused the `<NAME>__INSTALL_CONFIG_FILE_PACKAGE` option to not work as designed. This is because of incorrect quoting in the relevant CMake conditional statement in the implementation of `beman_install_library`.

* Adds detail the the `README.md` such that the existence of the `cmake/` subdirectory is documented. API documentation for `beman_install_library` was also added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,5 +9,49 @@ so it does not respect the usual structure of a Beman library repository nor The
 
 ## Description
 
+* `cmake/`: CMake modules and toolchain files used by Beman libraries.
 * `containers/`: Containers used for CI builds and tests in the Beman org.
 * `tools/`: Tools used to manage the infrastructure and the codebase (e.g., linting, formatting, etc.).
+
+## Usage
+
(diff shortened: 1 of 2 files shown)


bretbrownjr:
I'll be making a corresponding PR in `beman.exemplar` to update its `beman-submodule`, clean up the redundant lines that define `BEMAN_EXEMPLAR_INSTALL_CONFIG_FILE_PACKAGE`, and enhance its README in a couple places.

ednolan (approved):

nickelpro (approved):
Approval is for CMake changes.

On my side, pull the trigger if you don't have any other tweaks.

Issue #188, opened 2025-09-01
Requiring the R version of a paper is a maintenance burden


steve-downey:
The `implements` requirement seems to be checking for the R version number in the paper. This makes it very easy to get the revision number out of sync vs what is actually implemented. The wg21.link to an unversioned paper will automatically go to the last version.

neatudarius:
Agree. Eddie already change the standard here https://github.com/bemanproject/beman/pull/174.

@vickgoodman , since you implemented #49 , would you like do propagate this small update?

neatudarius:
We should make sure we support both versions of links: with and without R versions.
@vickgoodman , please add tests for them.

vickgoodman:
Updated here - #216 

Solved! Closing issue

PR #213, opened 2026-01-20
Added clearer comments to regex pattern

vickgoodman:
#188 - was already resolved. Just added clearer comments.

### DIFF
--- a/beman_tidy/lib/checks/beman_standard/readme.py
+++ b/beman_tidy/lib/checks/beman_standard/readme.py
@@ -114,21 +114,22 @@ class ReadmeImplementsCheck(ReadmeBaseCheck):
     def check(self):
         lines = self.read_lines_strip()
 
-        # Match the pattern to start with "Implements:" and then have a paper reference and a wg21.link URL.
+        # Match the pattern to start with "**Implements**:", then have a paper reference with a wg21.link URL.
         # Examples of valid lines:
-        # **Implements**: [Standard Library Concepts (P0898R3)](https://wg21.link/P0898R3).
-        # **Implements**: `std::ranges::any_view` proposed in [any_view (P3411)](https://wg21.link/p3411).
-        # **Implements**: [Give *std::optional* Range Support (P3168R2)](https://wg21.link/P3168R2) and [`std::optional<T&>` (P2988R5)](https://wg21.link/P2988R5)
-        # **Implements**: [.... (PxyzwRr)](https://wg21.link/PxyzwRr), [.... (PabcdRr)](https://wg21.link/PabcdRr), and [.... (PijklRr)](https://wg21.link/PijklRr),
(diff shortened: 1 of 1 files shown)


beman_tidy/lib/checks/beman_standard/readme.py
```diff
@@ -113,21 +114,22 @@ def __init__(self, repo_info, beman_standard_check_config):
...
 
-        # Match the pattern to start with "Implements:" and then have a paper reference and a wg21.link URL.
+        # Match the pattern to start with "**Implements**:", then have a paper reference with a wg21.link URL.
         # Examples of valid lines:
-        # **Implements**: [Standard Library Concepts (P0898R3)](https://wg21.link/P0898R3).
-        # **Implements**: `std::ranges::any_view` proposed in [any_view (P3411)](https://wg21.link/p3411).
-        # **Implements**: [Give *std::optional* Range Support (P3168R2)](https://wg21.link/P3168R2) and [`std::optional<T&>` (P2988R5)](https://wg21.link/P2988R5)
-        # **Implements**: [.... (PxyzwRr)](https://wg21.link/PxyzwRr), [.... (PabcdRr)](https://wg21.link/PabcdRr), and [.... (PijklRr)](https://wg21.link/PijklRr),
+        # **Implements**: [Standard Library Concepts (P0898)](https://wg21.link/P0898).
+        # **Implements**: `std::ranges::any_view` proposed in [any_view (P3411R4)](https://wg21.link/p3411r4).
```
neatudarius, line 120: I `p3411r4`(Lower case p and r) valid? (they should be but not seeing them in the regex). 

Please add unit tests for these 4 examples (I think we have tests with the same P paper copied in multiple tests, we can change that to use P+R, p+r and so one.

neatudarius (approved):
Changed my mind: do my suggestion in a follow-up PR.

We discussed in previous sync that we don't want to block PRs for extr changes not targetting the initial purpose

beman_tidy/lib/checks/beman_standard/readme.py
vickgoodman: p3411r4 (Lower case p and r) is valid for the wg21 link but not for the preceding label text. Do you want the text label to be uppercase only, or should I update the regex to support lowercase there too?
For example, should this be valid? (lowercase label text)
```
**Implements**: `std::ranges::any_view` proposed in [any_view (p3411r4)](https://wg21.link/p3411r4)
```

If looking at the Beman Standard for [readme.implements](https://github.com/bemanproject/beman/blob/main/docs/beman_standard.md#readmeimplements), nothing is specified but the label text is uppercase so please let me know how to proceed.

neatudarius: [p3411r4 ...](https://wg21.link/p3411r4) should be a valid thing

Issue #227, opened 2026-02-11
[beman-tidy] Add .beman-tidy.yml as a config file for ignoring certain paths


AndreiDurlea:
Example provided by @neatudarius :
```
ignored_paths:
    - include/beman/optional/detail/stl_interfaces/config.hpp
    - include/beman/optional/detail/stl_interfaces/fwd.hpp 
    - include/beman/optional/detail/stl_interfaces/iterator_interface.hpp
# or add support for regex
    - include/beman/optional/detail/stl_interfaces/ 
```

- [x] config file
- [x] ingoring paths
- [x] cli param to change default name
- [x] tests

neatudarius:
I think we should do this
1. Add .beman-tidy.yml as a generic config. For now, it can have only one entry
2. Always parse the config at startup and make sure:
2.1. We stop and prind error if ignored files include some mandatory paths in CI such as root README.md, root LICENSE, root
2.2 We ignore the paths in our get_*_files() utilities (should actually have a single funtion in utils/file.py where we extractc such files)
3. Tests and docs


Optionally:
4. Add a CLI param which can override the path to the config. Default value: root .beman-tidy

PR #228, opened 2026-02-15
Implemented a config file for beman-tidy

AndreiDurlea:
#227
Initial implementation:
- Load config with default name, `.beman-tidy.yml`, or `config_path` param.
- In the config yaml content, the entry `ignored_paths:` is used specify paths of dirs or files. Dirs exclude all files and dirs in them. (behaves like gitignore)
- Previously excluded dirs (.idea etc.) are now merged with this functionality in the `ignores` param found on `get_*_files()` functions.
- introduced `is_ignored()` function in `lib/utils/file.py` to check if a given path is ignored

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,30 +143,23 @@ beman-tidy path/to/exemplar --fix-inplace --verbose
 
 ## Configuration
 
-You can configure `beman-tidy` by placing a `.beman-tidy.yml` file in the root of your repository.
+`beman-tidy` attempts to read configuration for each source file from a `.beman-tidy.yml` file located in the root of your repository. You can also specify a custom configuration file path using the `--config` option.
 
-Currently, it supports ignoring specific paths or directories. When you add a path to `ignored_paths`, it will be excluded from all checks.
+The following configuration options may be used in a `.beman-tidy.yml` file:
 
-- To ignore a specific file, provide its full path relative to the repository root.
-- To ignore a directory, provide the path to that directory. This will ignore the directory itself and all files and subdirectories within it. A trailing slash (`/`) is optional.
+`ignored_paths` - A list of paths to be excluded from all checks.
+  - To ignore a specific file, provide its full path relative to the repository root.
+  - To ignore a directory, provide the path to that directory. This will ignore the directory itself and all files and subdirectories within it. A trailing slash (`/`) is optional.
 
-For example:
-```yaml
-ignored_paths:
-  - include/beman/optional/detail/stl_interfaces/config.hpp
-  - include/beman/optional/detail/stl_interfaces/
-```
-or
-```yaml
-ignored_paths:
-  - include/beman/optional/detail/stl_interfaces/
-```
-
-You can also specify a custom configuration file path using the `--config` option:
-
-```shell
-beman-tidy . --config my-config.yml
-```
+  Example:
+  ```yaml
+  ignored_paths:
+    # Ignores a single file
+    - include/beman/optional/detail/stl_interfaces/config.hpp
+    
+    # Ignores a directory and everything inside it
+    - include/beman/optional/another_dir
+  ```
 
 ## Working on beman-tidy
 
--- a/README.md
+++ b/README.md
@@ -145,11 +145,20 @@ beman-tidy path/to/exemplar --fix-inplace --verbose
 
 You can configure `beman-tidy` by placing a `.beman-tidy.yml` file in the root of your repository.
 
-Currently, it supports ignoring specific paths or directories:
+Currently, it supports ignoring specific paths or directories. When you add a path to `ignored_paths`, it will be excluded from all checks.
 
+- To ignore a specific file, provide its full path relative to the repository root.
+- To ignore a directory, provide the path to that directory. This will ignore the directory itself and all files and subdirectories within it. A trailing slash (`/`) is optional.
+
+For example:
 ```yaml
 ignored_paths:
   - include/beman/optional/detail/stl_interfaces/config.hpp
+  - include/beman/optional/detail/stl_interfaces/
+```
+or
+```yaml
+ignored_paths:
   - include/beman/optional/detail/stl_interfaces/
 ```
 
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ expedite adoption across the Beman Project.
 
 ```shell
 $ beman-tidy --help
-usage: beman-tidy [-h] [--fix-inplace | --no-fix-inplace] [--verbose | --no-verbose] [--require-all | --no-require-all] [--checks CHECKS] repo_path
+usage: beman-tidy [-h] [--fix-inplace | --no-fix-inplace] [--verbose | --no-verbose] [--require-all | --no-require-all] [--checks CHECKS] [--config CONFIG] repo_path
 
 positional arguments:
   repo_path             path to the repository to check
@@ -33,6 +33,7 @@ options:
   --require-all, --no-require-all
                         all checks are required regardless of the check type (e.g., Recommendation becomes Requirement)
   --checks CHECKS       array of checks to run
+  --config CONFIG       path to the configuration file (default: .beman-tidy.yml in repo root)
 ```
 
 - Run beman-tidy on the exemplar repository **(default: dry-run mode)**
@@ -138,6 +139,24 @@ Coverage          TOTAL:  95.83% (23/24 checks passed).
 
 ```shell
 beman-tidy path/to/exemplar --fix-inplace --verbose
+```
+
+## Configuration
+
+You can configure `beman-tidy` by placing a `.beman-tidy.yml` file in the root of your repository.
+
+Currently, it supports ignoring specific paths or directories:
+
+```yaml
+ignored_paths:
+  - include/beman/optional/detail/stl_interfaces/config.hpp
+  - include/beman/optional/detail/stl_interfaces/
+```
+
+You can also specify a custom configuration file path using the `--config` option:
+
+```shell
+beman-tidy . --config my-config.yml
 ```
 
 ## Working on beman-tidy
--- a/beman_tidy/.beman-tidy.yml
+++ b/beman_tidy/.beman-tidy.yml
@@ -1,0 +1,1 @@
+# Default beman-tidy configuration
--- a/beman_tidy/.beman-tidy.yml
+++ b/beman_tidy/.beman-tidy.yml
@@ -1,1 +1,3 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
 # Default beman-tidy configuration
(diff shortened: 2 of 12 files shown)


AndreiDurlea:
@neatudarius when you're free, can you take a look at this PR?

Besides the commited changes, I'm also thinking about a slight design improvement. Instead of forbidding, at config level, the ignoring of files/dirs such as `README.md`, `LICENSE` etc., I propose that their individual checks simply throw errors if those files _seem_ nonexistent and print out a log message if they actually exist but are ignored in the config. For example
`README.md was found, but was explicitly ignored in the config file (.beman-tidy.yml). Make sure it is visible to beman-tidy to pass this check and other related checks.`

I can easily implement something like this for any check that requires files to be present - that means modifying those checks to
1. Check in non-ignored files first
2. If some required file is not found, check the ignored stuff, just in case, and print out something useful (example above) if the mandatory file was ignored.

This would allow users to check for other issues while keeping something temporarily ignored for their own reasons.

And then I'd simply remove [`mandatory_files`](https://github.com/bemanproject/beman-tidy/pull/228/changes#diff-3ec8d50c30739e774012db2dd54f3db0f657b87694f3cfe22ad47a2bd132e68bR23) from `config.py`. Let me know if this design sounds ok

docs/dev-guide.md
```diff
@@ -13,6 +13,7 @@
 * `tests/`: Unit tests for the tool.
   * Structure is similar to the `beman_tidy/` directory.
   * `pytest` is used for testing.
+* `.beman-tidy.yml`: Configuration file for `beman-tidy` (optional).
```
neatudarius: Move it where ls -l says!

Should we first load default deployed `beman_tidy/.beman-tidy.yml`, and apply user file `repo/path/.beman-tidy.yml` over it? Aka should we provide a default .beman-tidy.yml near .beman-standard.yml? Right now it will be an empty config, but maybe we can add something in the future.

What do you think?

beman_tidy/lib/checks/base/file_base_check.py
```diff
@@ -3,17 +3,21 @@
...
+from ...utils.config import get_ignores
+
 
 class FileBaseCheck(BaseCheck):
     """
     Base class for checks that operate on a file.
     """
 
     def __init__(self, repo_info, beman_standard_check_config, relative_path, name=None):
+        self.repo_info = repo_info
```
neatudarius: Check ` super().__init__(repo_info, beman_standard_check_config, name=name)`. It is doing what you need, right?

beman_tidy/lib/checks/base/file_base_check.py
```diff
@@ -43,6 +47,29 @@ def pre_check(self):
...
 
+    def should_skip(self):
+        """
+        Check if the file should be skipped based on configuration.
+        """
+        if super().should_skip():
+            return True
+
+        ignores = get_ignores(self.repo_info)
+
```
neatudarius:
```suggestion
```
Please do not put empty lines without reasons. this connected with the next for, should be a compact block

beman_tidy/lib/checks/base/file_base_check.py
```diff
@@ -43,6 +47,29 @@ def pre_check(self):
...
+        rel_path_str = self.relative_path.as_posix()
+
```
neatudarius:
```suggestion
```

neatudarius (commented):
Initial round of review

docs/dev-guide.md
AndreiDurlea: That sounds like better design. Will work on implementing it 👍🏻

AndreiDurlea: [Done!](https://github.com/bemanproject/beman-tidy/pull/228/commits/54ca7c6537c867211ce441d333c966e7f7381786) Is it ok?

beman_tidy/lib/checks/base/file_base_check.py
AndreiDurlea: Yes, nice catch

AndreiDurlea: Okay so, after some investigative work:
- `repo_info` is (was) not available at init
- when I implemented `should_skip()` which calls `get_ignroes(self.repo_info)`which requires, of course, `repo_info`, a crash would happen
- I half-fixed that problem by just setting `repo_info` myself

I'm now working on a bugfix regarding this issue, by changing when some params are intialized in `base_check`'s constructor. Should I open a separate issue and do this in a separate pull request @neatudarius? Or is it fine to do this here.

AndreiDurlea:
@neatudarius I applied your changes, ready for another round of review.
Mention: I'm working on the deeper bugfix to remove [the temporary fix provided by L19 in file_base_check.py](https://github.com/bemanproject/beman-tidy/pull/228/changes#diff-f79f15f4db56c998457771c1b845d1d6c2c41ab811a221e46a86a7db7a2c1faaR19). Awaiting your response in the comment thread on that line.

README.md
```diff
@@ -140,6 +141,24 @@ Coverage          TOTAL:  95.83% (23/24 checks passed).
 beman-tidy path/to/exemplar --fix-inplace --verbose
 ```
 
+## Configuration
+
+You can configure `beman-tidy` by placing a `.beman-tidy.yml` file in the root of your repository.
+
+Currently, it supports ignoring specific paths or directories:
```
neatudarius: Let's change the formatting to this one: https://clang.llvm.org/extra/clang-tidy/

Image

README.md
```diff
@@ -140,6 +141,24 @@ Coverage          TOTAL:  95.83% (23/24 checks passed).
...
+
+```yaml
+ignored_paths:
```
neatudarius: if both examples would work, should be included in docs here

```
ignored_paths:
  - include/beman/optional/detail/stl_interfaces/config.hpp
  - include/beman/optional/detail/stl_interfaces/
 ```

or

```
ignored_paths:
  - include/beman/optional/detail/stl_interfaces/
```

File: beman_tidy/.beman-tidy.yml [hunk 7]
```diff
@@ -0,0 +1 @@
+# Default beman-tidy configuration
```
neatudarius, line 3: Let's add `# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception` (or equivalent) in each file

beman_tidy/lib/checks/base/file_base_check.py
neatudarius: I still don't understand why current status from main is not enough?

* repo info is set in super here : https://github.com/bemanproject/beman-tidy/blob/main/beman_tidy/lib/checks/base/base_check.py#L79

* path is et at this line https://github.com/bemanproject/beman-tidy/pull/228/changes#diff-f79f15f4db56c998457771c1b845d1d6c2c41ab811a221e46a86a7db7a2c1faaR24. if you still need `relative_path`, sure, add `self.relative_path = Path(relative_path)` at line 23.

TLDR: first line in any class that is derived needs to be called to super!

neatudarius (changes requested):

beman_tidy/lib/checks/base/file_base_check.py
AndreiDurlea: I managed to implement it so the first line is the super's constructor. See the changes [here](https://github.com/bemanproject/beman-tidy/pull/228/commits/70c3d6ef5d4346aedbbbfd1f7276e0fad0b896d0).

I initially didn't keep the constructor first because it needed `should_skip()` to set `log_level`. and my `should_skip()` needed properties that were not yet initialized (`repo_info`, `relative_path`) to find out if the check if for ignored paths. So I initialized them before the construcgtor and called it a day=)

Basically,
constructor -> `log_level` -> `should_skip()` -> stuff in the constructor. 

So i made `log_level` be lazily initialized.

`repo_info` was easy to fix, as I just made sure it was initialized before `log_level` in the constructor. But `relative_path` was not intiliazed in the constructor at all, and, *as you said*, the constructor should be the first thing called, hence the lazy initialization required for anything that uses `should_skip()` (in this case `log_level`

This issue had me use more brainpower than I expected I'd need for it =)))

beman_tidy/.beman-tidy.yml
AndreiDurlea: 08a8d9eab2583b5f3d66d25710397c2c0eee557c

README.md
AndreiDurlea: dca19c107f3925acf39b8a2ccf2391c720c587ad

README.md
AndreiDurlea: b13dd9dc272ec81d3a2abba75011a48afcfc0254

AndreiDurlea:
@neatudarius ready for a re-review

beman_tidy/lib/checks/base/directory_base_check.py
```diff
@@ -44,6 +46,14 @@ def pre_check(self):
 
         return True
 
+    def should_skip(self):
+        """
+        Check if the file should be skipped based on configuration.
+        """
+        if super().should_skip():
+            return True
+        return is_ignored(self.repo_info, self.relative_path)
```
neatudarius, line 55: Excellent!

beman_tidy/lib/utils/config.py
```diff
@@ -0,0 +1,111 @@
...
+             
+    return True
+
+
+def load_repo_config(repo_path, config_path=None):
+    """
+    Load the configuration file.
+    """
+    # Load default configuration
+    default_config_path = files('beman_tidy').joinpath('.beman-tidy.yml')
```
neatudarius: This is not good practice. Do you want something like https://github.com/bemanproject/beman-tidy/blob/main/beman_tidy/lib/utils/git.py#L125?

AndreiDurlea: Alright I get why, I'll implement that. I initially thought having that .parent.parent.parent structure would be prone to errors (for example if the file gets moved), but I suppose that's an okay compromise

neatudarius: Please test the installed tool. Does it work in your approach?

AndreiDurlea: It did run on exemplar, but, for the sake of consistency, I'll refactor to what is also used for `beman-standard.yml`. I dont see my file moving in the near future but, to be sure, I wrapped getting the path in a different function that will be sure to pop in someones eyes, should they want to move that file. Mental gymnastics here trying to prevent future bugs :)

[Done](https://github.com/bemanproject/beman-tidy/pull/228/commits/b8cb76046f935c036e5ab99ec30f60997143947b)! @neatudarius is ok?

AndreiDurlea:
@neatudarius yet again ready for a re-review

beman_tidy/lib/checks/base/file_base_check.py
neatudarius: Thanks! It looks good! :D

beman_tidy/lib/utils/config.py
neatudarius: yes

neatudarius (approved):
LGTM! @AndreiDurlea , thanks for your first big contribution in Beman! :D

AndreiDurlea:
Thanks! Has been a blast! Happy to merge this now :)

PR #240, opened 2026-04-20
Add skip for exemplar for readme.title check

vickgoodman:
Add skip for exemplar repo for readme.title check

### DIFF
--- a/beman_tidy/lib/checks/beman_standard/readme.py
+++ b/beman_tidy/lib/checks/beman_standard/readme.py
@@ -41,6 +41,9 @@ class ReadmeTitleCheck(ReadmeBaseCheck):
         super().__init__(repo_info, beman_standard_check_config)
 
     def check(self):
+        if self.short_name == "exemplar":
+            return True  # Skip the check for exemplar library, as it is a special case.
+
         lines = self.read_lines_strip()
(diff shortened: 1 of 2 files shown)


neatudarius (approved):

vickgoodman:
I reverted the changes on the README. It was supposed to be a different branch but messed it up. My mistake

How should I handle the pytest for readme.title for invalid and fix inplace because now they're failing.
Should I mark it with skip, smth like this: `@pytest.mark.skip(reason="...")`

neatudarius (approved):

beman_tidy/lib/checks/beman_standard/readme.py
```diff
@@ -41,6 +41,9 @@ def __init__(self, repo_info, beman_standard_check_config):
         super().__init__(repo_info, beman_standard_check_config)
 
     def check(self):
+        if self.short_name == "exemplar":
```
neatudarius, line 44: I think I better understood the scenario for exemplar after talk with @ednolan . We should discard this PR and proceed like described in https://github.com/bemanproject/beman-tidy/issues/266

neatudarius (changes requested):

PR #270, opened 2026-04-23
feat(cli): add --version flag (#269)

### LINKED ISSUE #269: [beman-tidy] Add --version option
## Goal

Add a `--version` flag to `beman-tidy` CLI so users can quickly check which version is installed.

## Motivation

Currently there is no way to check the installed version from the CLI. Example of current broken behavior:
```shell
$ pipx list | grep beman-tidy
   package beman-tidy 0.1.0, installed using Python 3.13.4
    - beman-tidy (symlink missing or pointing to unexpected location)

$ beman-tidy --version       # should work, but does not
usage: beman-tidy [-h] [--fix-inplace | --no-fix-inplace] [--verbose | --no-verbose] [--require-all | --no-require-all] [--checks CHECKS] [--config CONFIG] repo_path
beman-tidy: error: the following arguments are required: repo_path
```

This makes it hard to debug issues (e.g., user has `0.1.0` installed via `pipx` but expects `0.3.1` behavior). User has to run `pipx list | grep beman-tidy` instead.

## Expected behavior

```shell
$ beman-tidy --version
beman-tidy 0.4.0
```

## Implementation hint

Version is already defined in `pyproject.toml`. It can be read at runtime via:
```python
from importlib.metadata import version
print(version("beman-tidy"))
```
Check https://docs.python.org/3/library/importlib.metadata.html

Add `--version` to the `argparse` parser in `beman_tidy/cli.py`:
```python
parser.add_argument("--version", action="version", version=f"beman-tidy {version('beman-tidy')}")
```

mvanhorn:
Closes #269.

Adds a `--version` flag to the `beman-tidy` CLI so users can check the installed version without running `pipx list | grep beman-tidy`.

## Before

\`\`\`
\$ beman-tidy --version
usage: beman-tidy [-h] [--fix-inplace | --no-fix-inplace] ...
beman-tidy: error: the following arguments are required: repo_path
\`\`\`

## After

\`\`\`
\$ beman-tidy --version
beman-tidy 0.4.0
\`\`\`

## Implementation notes

- Uses argparse's built-in \`action="version"\`, which short-circuits before argparse validates the required \`repo_path\` positional.
- Reads the version via \`importlib.metadata.version("beman-tidy")\` rather than hardcoding a string — stays in sync with \`pyproject.toml\` automatically.
- \`importlib.metadata\` is stdlib; no new dependencies (project already requires Python 3.12+).

Ran \`uv run pytest tests/\` (69 passed) and \`uv run ruff check beman_tidy/cli.py\` (passes). Formatted with \`uv run ruff format\`.

### DIFF
--- a/beman_tidy/cli.py
+++ b/beman_tidy/cli.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 import argparse
+from importlib.metadata import version as _pkg_version
 import sys
 
 from beman_tidy.lib.utils.git import get_repo_info, load_beman_standard_config
@@ -14,6 +15,11 @@ def parse_args():
     """
 
     parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--version",
+        action="version",
+        version=f"beman-tidy {_pkg_version('beman-tidy')}",
+    )
     parser.add_argument("repo_path", help="path to the repository to check", type=str)
     parser.add_argument(
         "--fix-inplace",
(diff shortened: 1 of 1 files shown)


neatudarius (approved):
LGTM

PR #290, opened 2026-04-27
Implement check cmake.library name

vickgoodman:
- Implement cmake.library_name - #77 
- Add pytest tests

I thought I'd give the cmake.* checks a shot.

Implementation details:
- used [cmake_parser](https://roehling.github.io/cmake_parser/api/parser.html) that parses the CMakeLists.txt file and returns an AST
- from there the check was pretty straight forward


vickgoodman:
Tests are failing from library.name, not cmake.library_name..

vickgoodman:
> Tests are failing from library.name, not cmake.library_name..

Fixed.

It was failing because the `find_pytest_test_functions_for_check` function matched the cmake.library_name and library.name in the same way, so instead of counting them separate it would count them together

neatudarius (commented):
Thanks for the PR! The approach using `cmake_parser` for AST-based CMake parsing is the right direction — much more robust than regex for multi-line commands and avoids false positives from comments/strings.

A few things to fix before merging:

---

**🔴 Bug: `fix()` returns `None`**

```python
def fix(self):
    self.log("Please update the CMakeLists.txt file ...")
    # missing: return True
```

`fix()` must return `True` or `False`. Currently returns `None` (implicitly), which beman-tidy treats as a failed fix.

---

**🔴 Bug: the two invalid test files are identical**

`invalid-CMakeLists-v1` and `invalid-CMakeLists-v2` have exactly the same content. The comments say v1 = "CMakeLists missing library name" and v2 = "CMakeLists with invalid library name" — but both files show the same scenario (`add_library(beman::exemplar ALIAS beman.exemplar)` without a real target). Please differentiate them. Example of diff:
```
 $ git diff tests
diff --git a/tests/lib/checks/beman_standard/cmake/data/invalid/invalid-CMakeLists-v1 b/tests/lib/checks/beman_standard/cmake/data/invalid/invalid-CMakeLists-v1
index 4dc193c..8adae82 100644
--- a/tests/lib/checks/beman_standard/cmake/data/invalid/invalid-CMakeLists-v1
+++ b/tests/lib/checks/beman_standard/cmake/data/invalid/invalid-CMakeLists-v1
@@ -28,6 +28,7 @@ option(
 include(infra/cmake/beman-install-library.cmake)
 include(infra/cmake/BuildTelemetryConfig.cmake)
 
+# add_library(beman.exemplar INTERFACE)
 add_library(beman::exemplar ALIAS beman.exemplar)
 
 target_sources(
diff --git a/tests/lib/checks/beman_standard/cmake/data/invalid/invalid-CMakeLists-v2 b/tests/lib/checks/beman_standard/cmake/data/invalid/invalid-CMakeLists-v2
index 4dc193c..acf621c 100644
--- a/tests/lib/checks/beman_standard/cmake/data/invalid/invalid-CMakeLists-v2
+++ b/tests/lib/checks/beman_standard/cmake/data/invalid/invalid-CMakeLists-v2
@@ -28,6 +28,7 @@ option(
 include(infra/cmake/beman-install-library.cmake)
 include(infra/cmake/BuildTelemetryConfig.cmake)
 
+add_library(beman.exemplar2 INTERFACE)
 add_library(beman::exemplar ALIAS beman.exemplar)
 
 target_sources(
```

---

**🟡 Style: floating string literal — use `#` comments instead**

```python
def __init__(self, ...): ...

"""
parse_raw and parse_tree:
...
"""
def _get_cmake_parse_raw(self): ...
```

A triple-quoted string between methods is not a docstring — it's an unattached string literal. Use `#` comments here.

---

**🟡 `check()` logic can be simplified**

Current flow uses two separate `if` blocks that partially overlap. Cleaner:

```python
def check(self):
    ast = self._get_cmake_parse_raw()
    cmake_library_name = self._get_cmake_library_name(ast)

    if cmake_library_name is None:
        self.log("CMake library target name is not found. ...")
        return False

    if cmake_library_name != self.library_name:
        self.log(f"CMake library target name '{cmake_library_name}' does not match library name '{self.library_name}'. ...")
        return False

    return True
```

beman_tidy/lib/checks/beman_standard/cmake.py
```diff
@@ -10,9 +13,53 @@
...
+    - Hierarchical structures such as if() or function() blocks are not resolved.
+    
+    parsed_tree: 
+    - Returns a fully constructed AST.
+    - Resolves block structures such as function(), if(), include () and more.
+    - The block structures return specialized AST nodes (e.g., Function, If, Include, etc.).
+    
+    See more here: https://roehling.github.io/cmake_parser/api/parser.html
+    """
+    def _get_cmake_parse_raw(self):
```
neatudarius:
```suggestion
    def get_cmake_parse_raw(self):
```
We don't use the _ naming convention.

neatudarius (commented):
LGTM for current status. Let's move with this approach. It seems we may succeed to implement almost all cmake.* checks with cmake_parser

beman_tidy/lib/checks/beman_standard/cmake.py
```diff
@@ -27,6 +74,36 @@ def __init__(self, repo_info, beman_standard_check_config):
 
 
 # TODO cmake.library_name
```
neatudarius:
```suggestion
```

tests/lib/checks/beman_standard/cmake/data/invalid/invalid-CMakeLists-v1
neatudarius, line 1: I don't think we need full files for invalid scenarios.
Check

```
$ cat tests/lib/checks/beman_standard/cmake/data/invalid/invalid-library-name-v1.txt 
# ...
# missing add_library from next line
# add_library(beman.exemplar INTERFACE)
# ...

$ cat tests/lib/checks/beman_standard/cmake/data/invalid/invalid-library-name-v2.txt 
# ...
# wrong name: beman.exemplar2
add_library(beman.exemplar2 INTERFACE)
add_library(beman::exemplar ALIAS beman.exemplar)
# ...
                                                                                                                                                                                                                                              $ cat tests/lib/checks/beman_standard/cmake/data/invalid/invalid-library-name-v3.txt 
# ...
# wrong name: beman_exemplar
add_library(beman_exemplar INTERFACE)
add_library(beman::exemplar ALIAS beman.exemplar)
# ...
```

tests/lib/checks/beman_standard/cmake/data/valid/CMakeLists-v1.txt
neatudarius, line 1: This file is ok to be full.

I would rename it to CMakeLists-v1.txt, to match previous conventions (e.g., tests/lib/checks/beman_standard/readme/data/valid/README-v1.md)

neatudarius (approved):

PR #291, opened 2026-04-27
Implement skipped check for `cpp.extension_identifiers`

AndreiDurlea:
Skipped it as per discussed
#60

### DIFF
--- a/beman_tidy/lib/checks/beman_standard/cpp.py
+++ b/beman_tidy/lib/checks/beman_standard/cpp.py
@@ -111,6 +111,6 @@ class CppExtensionIdentifiersCheck(BaseCheck):
     def should_skip(self):
         self.log(
             "beman-tidy cannot actually check cpp.extension_identifiers. Please ensure that extension identifiers are prefixed with 'ext_'. "
-            "See https://github.com/bemanproject/beman/blob/main/docs/beman_standard.md#cppextensionidentifiers for more information."
+            "See https://github.com/bemanproject/beman/blob/main/docs/beman_standard.md#cppextension_identifiers for more information."
         )
         return True
--- a/beman_tidy/lib/checks/beman_standard/cpp.py
+++ b/beman_tidy/lib/checks/beman_standard/cpp.py
@@ -99,6 +99,17 @@ class CppNamespaceCheck(BatchFileBaseCheck):
             self.write_lines(new_lines)
             return True
 
+from beman_tidy.lib.checks.base.base_check import BaseCheck
+
 # TODO cpp.no_flag_forking
 
-# TODO cpp.extension_identifiers
+@register_beman_standard_check("cpp.extension_identifiers")
+class CppExtensionIdentifiersCheck(BaseCheck):
+    def __init__(self, repo_info, beman_standard_check_config):
+        super().__init__(repo_info, beman_standard_check_config)
+
+    def should_skip(self):
+        self.log(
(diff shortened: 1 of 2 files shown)


beman_tidy/lib/checks/beman_standard/cpp.py
```diff
@@ -99,6 +99,17 @@ def fix(self):
...
 
-# TODO cpp.extension_identifiers
+@register_beman_standard_check("cpp.extension_identifiers")
+class CppExtensionIdentifiersCheck(BaseCheck):
+    def __init__(self, repo_info, beman_standard_check_config):
+        super().__init__(repo_info, beman_standard_check_config)
+
+    def should_skip(self):
+        self.log(
+            "beman-tidy cannot accurately check cpp.extension_identifiers. Please ensure that extension identifiers are prefixed with 'ext_'."
```
neatudarius: Please use similar message to other skipped rules. Check the message for other examples.

I think other examples have tests

beman_tidy/lib/checks/beman_standard/cpp.py
```diff
@@ -99,6 +99,17 @@ def fix(self):
             self.write_lines(new_lines)
             return True
 
+from beman_tidy.lib.checks.base.base_check import BaseCheck
```
neatudarius: Import must be at the top of the file, not mid-file. Also, `BaseCheck` may already be available through existing imports — check if you can reuse what's already there.

beman_tidy/lib/checks/beman_standard/cpp.py
```diff
@@ -99,6 +99,17 @@ def fix(self):
...
+
 # TODO cpp.no_flag_forking
```
neatudarius, line 104: `cpp.no_flag_forking` appears before `cpp.extension_identifiers` in both the Beman Standard and `.beman-standard.yaml`. Please preserve that order — either move the implementation after the `no_flag_forking` TODO, or keep the TODO in place and append the implementation at the end of the file.

beman_tidy/lib/checks/beman_standard/cpp.py
```diff
@@ -99,6 +99,17 @@ def fix(self):
...
 
-# TODO cpp.extension_identifiers
+@register_beman_standard_check("cpp.extension_identifiers")
+class CppExtensionIdentifiersCheck(BaseCheck):
```
neatudarius, line 107: Every check registered with `@register_beman_standard_check` requires exactly 3 test functions: `test__cpp_extension_identifiers__valid`, `test__cpp_extension_identifiers__invalid`, `test__cpp_extension_identifiers__fix_inplace`. The registry validates this automatically — without them `test_registry.py` will fail. Please add a test file following the pattern in `tests/lib/checks/beman_standard/cpp/`.

neatudarius (commented):
Overall the `should_skip` pattern is correct and consistent with `readme.purpose`. See inline comments for a few things to fix before merging.

neatudarius (commented):
LGTM for current state. We want to skip this check!

beman_tidy/lib/checks/beman_standard/cpp.py
AndreiDurlea: Fixed test registry (it actually just asks for one func)
```python
# Skipped checks have exactly one test function.
expected_function_name = f"test__{check_pattern}__is_always_skipped"
```

neatudarius (approved):

PR #305, opened 2026-05-01
Implement file.test_names

AndreiDurlea:
#59 
This implements file.test_names for `--dry-run`. As for fix-inplace, renaming the files automatically would cause build system inconsistencies so I've decided to skip that.


beman_tidy/lib/checks/beman_standard/file.py
```diff
@@ -14,7 +14,46 @@
 # TODO file.names
 
 
-# TODO file.test_names
+@register_beman_standard_check("file.test_names")
+class FileTestNamesCheck(BatchFileBaseCheck):
+    """
+    [file.test_names]
+    Requirement: Test source code files must use the *.test.cpp naming convention.
```
neatudarius, line 65: Correct - this is a requirement https://github.com/bemanproject/beman/blob/main/docs/beman_standard.md#filetest_names (https://github.com/bemanproject/beman-tidy/blob/main/beman_tidy/.beman-standard.yaml#L131)

(no change requested here)

tests/lib/checks/beman_standard/file/test_file.py
neatudarius, line 1: Maybe we should also add `identity_test.cpp` as an invalid test case.

neatudarius (approved):
LGTM, but please check my comments before merge.

Issue #241, opened 2026-04-22
[bemanification] Run beman-tidy on beman.exemplar


neatudarius:
Goal: Add `beman-tidy` into CI precommit hooks for generate libraries from beman.examplar (that means only inside the `cookiecutter` directory - as discussed with @ednolan).

Note: We don't run `beman-tidy` on the actual examplar library.

CrepuscularIRIS:
I ran `beman-tidy` on `beman.exemplar` and found three violations:

- **[Requirement] directory.docs**: `CONTRIBUTING.md` at root is a misplaced MD file (should be in `docs/`)
- **[Requirement] directory.papers**: Same `CONTRIBUTING.md` triggers the paper-file check
- **[Recommendation] readme.title**: First line of `README.md` was `# How to Use This Template` instead of the required `# beman.exemplar: <description>`

I'm opening a PR on [bemanproject/exemplar](https://github.com/bemanproject/exemplar) to fix these. The fix:
1. Moves `CONTRIBUTING.md` → `docs/CONTRIBUTING.md` (GitHub also recognizes CONTRIBUTING.md in `docs/`)
2. Extracts the "How to Use This Template" preamble into `docs/howto.md`
3. Lets `README.md` start with `# beman.exemplar: A Beman Library Exemplar`

After the fix, `beman-tidy exemplar/` passes with 0 Requirement failures and 0 Recommendation failures.

neatudarius:
FYI @vickgoodman

vickgoodman:
Hello, please update the beman-tidy utilitary via `uv tool upgrade beman-tidy`. Let me know if that was the issue.

neatudarius:
@AndreiDurlea , is this ready?

AndreiDurlea:
Just reran CI https://github.com/bemanproject/beman-tidy/actions/runs/25210491553?pr=296 on #296. All green. I'd say it's ready, we can merge that and close this

neatudarius:
@AndreiDurlea  - check https://github.com/bemanproject/beman-tidy/actions/runs/25333715930/job/74273525809 (latest jobs on main branch)

AndreiDurlea:
Insane. Thought it was because of windows - and amazingly it ran on my machine with no failed checks after i fixed the windows bug. Will look into this asap. Might be another bug that I couldnt see locally.

AndreiDurlea:
Oh nvm. Found out why. Easy fix

[...]

PR #309, opened 2026-05-04
ci: make run-on-exemplar run on template output

AndreiDurlea:
fixes #241 (attempt 2!)
> Goal: Add beman-tidy into CI precommit hooks for generate libraries from beman.examplar (that means only inside the cookiecutter directory - as discussed with @ednolan).

PR #311, opened 2026-05-04
Make file.names ignore tests

### LINKED ISSUE #310: [bug] file.names makes files with ".test" fail
Should ignore test folder as there is file.test_names for that

AndreiDurlea:
Fixes #310


AndreiDurlea:
CI fails because cookie cutter folder gets checker before template generation - fixed in #309, that needs to be merged for this to be green

Once both are merged CI will be green on main and we can close #241

@neatudarius can we merge this?

neatudarius (approved):

PR #312, opened 2026-05-04
concept: make CI run on multiple platforms to test for subtler bugs

AndreiDurlea:
Updated workflows to include a test matrix (similar to what's on exemplar) that targets ubuntu, windows, macos.

### DIFF
--- a/.github/workflows/beman-tidy.yml
+++ b/.github/workflows/beman-tidy.yml
@@ -34,8 +34,15 @@ jobs:
           uv run ruff check --diff
 
   run_tests:
-    name: Run Tests
-    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


AndreiDurlea:
Hah, caught a problem right here and there. Windows users can't even use the stamp script without WSL (like I do). Will open a PR in exemplar to change stamp.sh to stamp.py

neatudarius:
@AndreiDurlea , do you plan to continue this PR?

AndreiDurlea:
Yeah! I actually added a new script in a PR on exemplar so we could test beman tidy on multiple platforms, so this PR is waiting for that one to get merged

PR #326, opened 2026-06-08
[beman-tidy] Implement cmake.target_names check

vickgoodman:
- Implement cmake.target_names check - #75
- Add unit tests

### DIFF
--- a/beman_tidy/lib/checks/base/base_check.py
+++ b/beman_tidy/lib/checks/base/base_check.py
@@ -77,6 +77,8 @@ class BaseCheck(ABC):
         assert self.repo_path is not None
         self.library_name = f"beman.{self.short_name}"
         assert self.library_name is not None
+        self.library_alias = f"beman::{self.short_name}"
+        assert self.library_alias is not None
 
         # set beman library maturity model
         beman_library_maturity_model = beman_standard_check_config[
--- a/beman_tidy/lib/checks/beman_standard/cmake.py
+++ b/beman_tidy/lib/checks/beman_standard/cmake.py
@@ -48,7 +48,8 @@ class CMakeBaseCheck(FileBaseCheck):
     def get_cmake_parse_tree(self, skip_comments=True):
         return cmake_parser.parser.parse_tree(self.read(), skip_comments=skip_comments)
 
-    def get_cmake_library_name(self, ast):
+    def get_cmake_library_name(self):
+        ast = self.get_cmake_parse_raw()
         cmake_library_name = None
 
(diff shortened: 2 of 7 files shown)


neatudarius (approved):
Thanks for implementing `cmake.target_names` — the logic correctly matches the standard and CI is green, so I'm approving to unblock.

**That said, please address the feedback below before merging** — especially point 1, which leaves the happy path untested.

**1. (Medium) The "valid" test never exercises the positive path the rule is about.**
`tests/.../cmake/data/valid/CMakeLists-v1.txt` only contains:
```cmake
add_library(beman.exemplar INTERFACE)
add_library(beman::exemplar ALIAS beman.exemplar)
```
Both are skipped by the `library_name`/`library_alias` guards, so the branch that validates the prefix (`target_name.startswith(f"{self.library_name}.")`) is **never executed** in the passing test — it passes vacuously. Since correctly-prefixed example/test targets are the whole point of the rule, a regression in the `startswith` branch would not be caught. Please add a valid fixture containing a properly-prefixed target, e.g. `add_executable(beman.exemplar.examples.basic)`.

**2. (Low, scope) The check only sees the top-level `CMakeLists.txt`.**
`get_cmake_target_names()` uses `parse_raw`, which does not resolve `add_subdirectory`. In real Beman repos the example/test targets this rule governs typically live in subdirectory CMakeLists files added via `add_subdirectory(...)`, so in practice the check will rarely see the targets it's meant to police and will silently pass. This is consistent with the other `cmake.*` checks and arguably out of scope here — but worth a TODO/comment or a follow-up issue noting the limitation.

**3. (Low) Misleading message for misnamed library targets.**
In `check()`, an `add_library` first-arg that isn't exactly `library_name`/`library_alias` is reported as `"Invalid CMake target name ... expected prefix: '<library_name>.'"`. For `invalid-target_names-v1.txt` (`add_library(beman.foobar)`) and `-v3.txt` (`add_library(beman.exemplar2 ...)`), the offending line is really the *library target itself*, which is the domain of `cmake.library_name`. Not a bug (still returns `False`), but the diagnostic conflates two rules. Invalid fixtures for this check would be more on-point using `add_executable` example/test targets with a bad prefix (v2 already does this well).

vickgoodman:
1. Done
2. Added a note and TODO to cmake.targe_names check
3. Kept the code as-is

vickgoodman:
@neatudarius 
Same failed workflows as mentioned here: https://github.com/bemanproject/beman-tidy/pull/325#issuecomment-4969479532

PR #343, opened 2026-07-15
[beman-tidy] Implement cmake.default check

neatudarius:
Implements #41


beman_tidy/lib/checks/beman_standard/cmake.py
```diff
@@ -127,7 +127,89 @@ def get_cmake_target_names(self):
...
+        super().__init__(repo_info, beman_standard_check_config)
+
+    def _is_library_add_library_command(self, command):
+        if command.identifier != "add_library":
+            return False
+        if not command.args:
+            return False
+        if command.args[0].value != self.library_name:
+            return False
+        if len(command.args) >= 2 and command.args[1].value == "ALIAS":
```
neatudarius:
```suggestion
        if len(command.args) >= 2 and command.args[1].value.upper() == "ALIAS":
```

